PR #6, opened 2018-09-05
Call the app `new_relic_agent`

binaryseed:
The name `new_relic` is already taken, so we already call the Hex package `new_relic_agent`. This PR switches the name of the application as well, so we don't get confused.


mattbaker:
🚢

PR #13, opened 2018-09-07
Use HTTPoison for agent http requests

binaryseed:
This PR uses `HTTPoison` for internal agent http requests, which also means it's now a non-optional dep.

@mattbaker


mattbaker:
🐑 

what motivated this?

PR #19, opened 2018-09-19
Report as elixir

binaryseed:
Update the agent to report as `elixir` instead of `sdk`

**WIP**: the New Relic ingest pipeline is being updated to support this


stefanahman:
@binaryseed still wip?

binaryseed:
Yeah, the work is actually on the New Relic server side... We'll get there :)

PR #31, opened 2018-10-01
Extend the timeout on harvester send_harvest

binaryseed:
This PR extends the time that a harvester Task is willing to wait for the `send_harvest` action. The action is essentially doing an HTTPoison request, which has it's own built-in timeout values of `8000` for connecting and `5000` for receiving a response.

The Task isn't blocking anything so this won't impact anything else in the agent.

### DIFF
--- a/lib/new_relic/harvest/collector/custom_event/harvester.ex
+++ b/lib/new_relic/harvest/collector/custom_event/harvester.ex
@@ -47,15 +47,6 @@ defmodule NewRelic.Harvest.Collector.CustomEvent.Harvester do
       |> Collector.HarvestCycle.current_harvester()
       |> GenServer.call(:gather_harvest)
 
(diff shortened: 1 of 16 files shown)


lib/new_relic/harvest/collector/custom_event/harvester.ex
```diff
@@ -51,7 +51,7 @@ defmodule NewRelic.Harvest.Collector.CustomEvent.Harvester do
 
   def complete(harvester) do
     Task.Supervisor.start_child(Collector.CustomEvent.TaskSupervisor, fn ->
-      GenServer.call(harvester, :send_harvest)
+      GenServer.call(harvester, :send_harvest, 15_000)
```
sb8244: Is it worth extracting this into a module function so that all collectors utilize the same timeout?

PR #36, opened 2018-10-04
Optionally disable error reporting

sb8244:
For environments where error reporting is not desired, the `error_reporting_enabled` config can be
set to false.

@binaryseed do you have any suggestions on the best way to test the dynamic application.ex setup? I was thinking of an integration style test in transaction_error_event_test.exs, but it would involve tearing down the entire application supervisor and restarting it. That could be done with `async: false` and would be safe. edit: Turns out that async: false is the default, so it's simply a matter of wanting to go that route.


sb8244:
I think that this is fully tested now. I went with an integration style test that kills the entire application with the updated config (and removes the SASL logger), as well as a test that doesn't require killing the application.

PR #38, opened 2018-10-08
Extend Beam sampler stats

binaryseed:
This PR adds a few more statistics to the `Beam` sampler, including scheduler utilization

Also changed variable names since I got confused. `previous` is better than `last` since it's chronological order we are comparing.


binaryseed:
well I found this..

http://erlang.org/doc/man/scheduler.html#utilization-2

```erlang
one = :scheduler.sample()
two = :scheduler.sample()
:scheduler.utilization(one, two)
```

which seems to do the hard part for us

mattbaker:
very cool

PR #41, opened 2018-10-10
Detect and sample the top processes

binaryseed:
This PR adds a new sampler that will detect the top processes (according to memory & message queue length) and report on them

Also
* A few tweaks to the other process sampler to make it safer
* Randomize the start of each of the samplers so they all don't run at once


binaryseed:
Had to rebase @mattbaker

sb8244:
@binaryseed where will these show up in New Relic? Really love this feature but I'm not sure how I'd find the info in NR.

binaryseed:
It's all based on Insights events, so right now it's visible via dashboards.

However, I'm working on a BEAM page for elixir apps, which will display all this inside APM :)

PR #45, opened 2018-10-12
Properly format top process names

binaryseed:
Inspecting a module gets rid of the `Elixir.` module prefix


binaryseed:
Fixed with https://github.com/newrelic/elixir_agent/commit/957fe2db9f249002aae761e915bc37c64106b7a0

PR #49, opened 2018-10-15
Identify Phoenix transactions using the controller and action

sb8244:
This will create a transaction named like `/Phoenix/MyAppWeb.MyController/show` based on the
controller and action name.

It has been discussed putting this in a separate  package. However, I do believe it to be so common that it is best served in the base package.


binaryseed:
Thanks for doing this! I also found some time and made a small package for this:

https://github.com/binaryseed/new_relic_phoenix

I'm going to close this but thanks for putting in the time, I like how you tested it..

Issue #10, opened 2018-09-06
Cowboy 2.x requirement


fbjork:
Any plans on supporting Cowboy 1.0? I think that would increase adoption of Phoenix based apps.

binaryseed:
Hi, we leverage a feature in Cowboy 2.0 that ensures that each HTTP request is handled in it's own process.. w/o that, strange things happen when trying to track what transaction any given work was a part of...

I don't use Phoenix myself, but it looks like Phoenix `master` now has support for cowboy 2, so it's likely that support is right around the corner...

fbjork:
@binaryseed Makes sense. Hopefully Phoenix 1.4 lands soon.

Are you planning on supporting Phoenix and Ecto in this library? If so will you accept PRs?

binaryseed:
We're open to framework instrumentation beyond `Plug`, but we'd like to go the direction of making this project extensible so that there could be a package (something like `new_relic_phoenix` where that logic could live, and it could post data into this newrelic application. Mind opening an issue for those so we can track them as feature requests?

fbjork:
@binaryseed sure thing, just created a separate issue for a Phoenix library.

binaryseed:
I'll keep this open until Phoenix supports cowboy 2 out of the box..

sb8244:
What would the error be like if overridden? Would it fail to report that transaction, or something worse?

binaryseed:
Yes, it'd be a failure to report some data, it wouldn't impact the request at all. All reporting is done in a supervision tree separate from your running application.

binaryseed:
That actually looks like the agent timed out trying to send data to New Relic, which I think would be unrelated to this issue.

binaryseed:
FYI, Looks like a `rc` of 1.4 is out:

https://github.com/phoenixframework/phoenix/blob/master/CHANGELOG.md#140-rc0-2018-10-09

fbjork:
Closing since Phoenix 1.4 is around the corner.

PR #62, opened 2018-11-26
Set cowboy dependency to be test only

mmrobins:
This dependency blocks usage for phoenix apps on 1.3 since they have a
requirement for cowboy 1.x.  This shouldn't be necessary since cowboy is
only used in tests.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -38,7 +38,7 @@ defmodule NewRelic.Mixfile do
       {:ex_doc, "~> 0.19", only: :dev, runtime: false},
       {:jason, "~> 1.0"},
       {:plug, "~> 1.6"},
-      {:cowboy, "~> 2.0"},
+      {:cowboy, "~> 2.0", only: :test},
       {:httpoison, "~> 1.0"}
     ]
   end


binaryseed:
You should be able to use the `override: true` flag to do this... We do actually rely on `plug` and `cowboy 2` in the agent

mmrobins:
Hmm, I guess I assumed cowboy wasn't used since the only place I see it referenced other than tests is a method name ```
╭─[~/work/new_relic_elixir_agent]─[mrobinson@C02SX132H03Y]─[20181126 - 15:00:42]-‹cowboy_just_test›
╰─ [0]% ack -i cowboy lib
lib/new_relic/transaction/reporter.ex
217:      cowboy_process_event(tx_attrs, pid)
253:  defp cowboy_process_event(tx_attrs, pid) do
260:      name: "Cowboy Process #{inspect(pid)}",
```

and I was able to get a transaction trace to report from phoenix using my branch 
<img width="1613" alt="image" src="https://user-images.githubusercontent.com/9961/49047472-6363cb00-f18c-11e8-926b-26ad22af36d4.png">

I'm confused why something that in the end uses httpoison to send an http request would need a web server like cowboy in the code, but I admit I haven't fully grokked this codebase yet.

binaryseed:
The agent provides a plug to do transaction tracing. We require that plug is run with Cowboy 2 ( see #10 for why ) - it doesn't call the code directly but it relies on Cowboy 2 being used not Cowboy 1, which is why we list the dependency...

mmrobins:
Great, the info on https://github.com/newrelic/elixir_agent/issues/10 is very helpful for explaining things, thanks for point that out.

PR #68, opened 2018-12-02
Enable communication through a proxy

binaryseed:
This PR adds support for communicating through a proxy. 

Opt-in via config:
```elixir
config :new_relic_agent, proxy_url: "http://my.proxy:8080"
```

closes #60

### DIFF
--- a/lib/new_relic/config.ex
+++ b/lib/new_relic/config.ex
@@ -34,6 +34,12 @@ defmodule NewRelic.Config do
     do: System.get_env("NEW_RELIC_LOG") || Application.get_env(:new_relic_agent, :log)
 
   @doc """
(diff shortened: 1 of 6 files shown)


binaryseed:
I'm actually going to skip this in favor of switching to erlang `:httpc` which can be configured separate from the agent

lenfree:
hi @binaryseed , could you point me to the doco how to configure proxy env var with `:httpc`?

binaryseed:
http://erlang.org/doc/apps/inets/http_client.html

lenfree:
> http://erlang.org/doc/apps/inets/http_client.html

Sorry I am new to Elixir. I had a look at the doco, but I couldn't find a way to set this via environment variable or via config like this? 

```
config: new_relic_agent,
  http_options: [proxy: "http://localhost:3128"]
```

PR #70, opened 2018-12-04
Switch to :httpc

binaryseed:
This PR switches the agent to use the built-in `:httpc` library for HTTP requests
* This removes a handful of dependencies which is nice
* Also included is a Mozilla certificate chain for SSL verification
* `gzip` the request body, which can reduce the size by 3-4x

sidekick/ @mattbaker


lib/new_relic/util/ssl.ex
```diff
@@ -0,0 +1,21 @@
...
+  @external_resource @cacertfile
+  @cacerts File.read!(@cacertfile)
+           |> :public_key.pem_decode()
+           |> Enum.map(fn {:Certificate, der, _} -> der end)
+
+  def ssl_options() do
+    [
+      ssl: [
+        verify: :verify_peer,
+        cacerts: @cacerts,
```
voltone: I would put cacert.pem in the package's /priv directory and pass `cacertfile` instead. Passing certificates/keys as DER binaries instead of paths is not tested very well, and there have been issues with it in past OTP releases. The caching by the ssl application works quite well, so there is no negative impact.

lib/new_relic/util/ssl.ex
```diff
@@ -0,0 +1,21 @@
...
+        # See: https://github.com/erlang/otp/blob/8ca061c3006ad69c2a8d1c835d0d678438966dfc/lib/ssl/src/ssl.erl#L1121
+        verify_fun: fn [{:bad_cert, _}] -> false end
```
voltone: Why the custom verify_fun? This overrides the default, which takes care of things like hostname verification. If you keep the default `verify_fun` you get automatic hostname verification when on OTP 20 or later.

If you want to support older OTP versions you'd have to pass in your own `verify_fun`, but it would have to be more elaborate than that. Easiest would probably be to use `ssl_verify_fun`, and use it regardless of OTP version in use.

That would also take care of another issue: the built in hostname verification of OTP >=20 does not handle wildcard certs with Subject Alternative Name extensions. On OTP >=21 you could pass `{customize_hostname_check, [{match_fun, public_key:pkix_verify_hostname_match_fun(https)}]}` to fix that.

All in all, probably best to stick to `ssl_verify_fun`. That does require extracting the hostname from the URL, though.

binaryseed: Thanks for this info, I went this route not because I'm sure it's right, but it's what I got to work. With no `verify_fun` I simply couldn't get a request to work. I did this to get this bit of code to trigger:

https://github.com/erlang/otp/blob/8ca061c3006ad69c2a8d1c835d0d678438966dfc/lib/ssl/src/ssl.erl#L1121-L1135

```erlang
validate_option(verify_fun, Fun) when is_function(Fun) ->
    {fun(_,{bad_cert, _} = Reason, OldFun) ->
	     case OldFun([Reason]) of
		 true ->
		     {valid, OldFun};
		 false ->
		     {fail, Reason}
	     end;
	(_,{extension, _}, UserState) ->
	     {unknown, UserState};
	(_, valid, UserState) ->
	     {valid, UserState};
	(_, valid_peer, UserState) ->
	     {valid, UserState}
     end, Fun};
```

I'll update this to just use `ssl_verify_fun`

That being said, given these complexities, and the comments over in #60 , the decision of wether or not to go this route or to wait for `xhttp` is still up in the air

lib/new_relic/util/ssl.ex
binaryseed: That's a good idea, I'll do that. I got this technique from the code in Hex but this does seem safer.

lib/new_relic/util/ssl.ex
voltone: The reason it didn't work at first must be the wildcard cert with SAN at collector.newrelic.com:

```
Subject: C=US, ST=California, L=San Francisco, O=New Relic, Inc., CN=*.newrelic.com
[...]
    X509v3 Subject Alternative Name: 
        DNS:*.newrelic.com, DNS:newrelic.com
```

With your `verify_fun`, the URL's hostname was not being checked against the certificate at all, so it started working, but of course that allows impersonation and MitM attacks.

On OTP 21 you can make it work with the `customize_hostname_check` option, but I imagine you'll want to support older versions as well...

binaryseed: Thanks I'm not versed in the details of certificates so this is really helpful!!!!

binaryseed:
@josevalim do you think it's a reasonable choice to leverage `:httpc` temporarily until `xhttp` is released?

The concerns around load are low risk since the agent only makes a handful of requests per minute. And I believe I've addressed SSL concerns by leveraging the `ssl_verify_fun` library as advised by @voltone

PR #78, opened 2019-01-09
Add another parsing function for exception shapes

binaryseed:
This PR cleans up the shape of some kinds of exceptions that currently are wrapped up in a few layers. This directly extracts the relevant exception & stacktrace


mattbaker:
💯

PR #84, opened 2019-01-21
Add basic support for "Other" transactions

### LINKED ISSUE #83: Non-web transactions tracing
**Is your feature request related to a problem? Please describe.**
I would like to have Non-web transactions represented in NewRelic transactions list. As an example, I would like to see [ConsumerSupervisor](https://hexdocs.pm/gen_stage/ConsumerSupervisor.html) children processes as transactions.

**Describe the solution you'd like**
It would be great to have module `NewRelic.NonWebTransaction` which I could use in my module and then annotate function, which I would like to create transaction for (similar to NewRelic.Trace, but created transaction instead of trace span).

I would be eager to help implementing it.

**Describe alternatives you've considered**
I'm wrapping my functions with something like:
```elixir
def new_relic_transaction(name, fun) do
  NewRelic.Transaction.Reporter.start()
  NewRelic.set_transaction_name(name)
  fun.()
after
  conn = %Plug.Conn{method: "NONWEB"} # Fake connection
  NewRelic..Transaction.Reporter.stop(conn)
end
```

But it does report transactions as `Web`.

binaryseed:
This PR enables tracking a given process as an "Other" transaction (ie: not a Web transaction). Had to do a minimal amount of refactoring to allow this, and included a little re-shuffling of functions.

```elixir
NewRelic.start_transaction("GenStage", "FancyConsumer/EventType")
```

You may track a process that isn't currently part of another transaction. (note: A process spawned in a Web request _is_ part of that web transaction). 

Be careful tracking long running processes, as memory usage will grow while collecting attributes & function traces

Normal Web transactions should still `use NewRelic.Transaction`

closes #83


barthez:
What about distributed trace? Is `NewRelic.start_transaction("GenStage", "FancyConsumer/EventType")` going to create distributed trace context so all instrumented `HTTPoison` requests are going to be recorded properly?

PR #85, opened 2019-01-21
Disallow altering framework_name attribute

barthez:
Add "immutable attributes" list, which defines transaction attributes that can be set only once and cannot be overwritten.

The `framework_name` attribute is an example, as the first framework (deepest in the callstack) sets the transaction name, it should also sign off the transaction.

Closes https://github.com/binaryseed/new_relic_absinthe/issues/4


binaryseed:
Thanks for the submission, but this won't solve the problem you expressed, please see my comment on https://github.com/binaryseed/new_relic_absinthe/issues/4

PR #89, opened 2019-01-29
Safer AWS metadata query

### LINKED ISSUE #74: New Relic agent supported for services running within AWS Fargate?
Hi there,

we don't seem to be able to run the agent from our Elixir service running from AWS Fargate.

The AWS metadata endpoint:
https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/util.ex#L68

is not available from an Fargate/ECS task as per this forum post:
https://forums.aws.amazon.com/thread.jspa?messageID=830807&tstart=0

and we are getting this error upon start of the Elixir service:

```
Elixir.ArgumentError argument error 
    gen_tcp.erl:153 :gen_tcp.connect/4
    APP_PATH/deps/hackney/src/hackney_connect.erl:247 :hackney_connect.do_connect/5
    APP_PATH/deps/hackney/src/hackney_connect.erl:37 :hackney_connect.connect/5
    APP_PATH/deps/hackney/src/hackney.erl:316 :hackney.request/5
    lib/httpoison/base.ex:630 HTTPoison.Base.request/9
    lib/new_relic/util.ex:80 NewRelic.Util.aws_vendor_hash/1
    lib/new_relic/util.ex:71 NewRelic.Util.maybe_add_vendors/2
    lib/new_relic/harvest/collector/agent_run.ex:62 NewRelic.Harvest.Collector.AgentRun.connect_payload/0
```

Is Fargate not supported yet or are we missing anything?

Thanks!
Angel

**Environment**
```
Erlang/OTP 21 [erts-10.0.5] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:1] [hipe]

Elixir 1.7.2 (compiled with Erlang/OTP 21)
```
* Agent version 1.3.1

binaryseed:
Closes #74 by protecting against exceptions when attempting to see if AWS metadata is accessible.

Most of this PR is a re-shuffle of the vendor specific functions

### DIFF
--- a/lib/new_relic/util/vendor.ex
+++ b/lib/new_relic/util/vendor.ex
@@ -1,0 +1,44 @@
+defmodule NewRelic.Util.Vendor do
+  def maybe_add_linux_boot_id(util) do
+    case File.read("/proc/sys/kernel/random/boot_id") do
+      {:ok, boot_id} -> Map.put(util, "boot_id", boot_id)
+      _ -> util
+    end
+  end
+
+  def maybe_heroku_dyno_hostname do
+    System.get_env("DYNO")
+    |> case do
+      nil -> nil
+      "scheduler." <> _ -> "scheduler.*"
+      "run." <> _ -> "run.*"
+      name -> name
+    end
+  end
+
+  @aws_url "http://169.254.169.254/2016-09-02/dynamic/instance-identity/document"
+  def maybe_add_cloud_vendors(util, options \\ []) do
+    Keyword.get(options, :aws_url, @aws_url)
+    |> aws_vendor_hash()
+    |> case do
+      nil -> util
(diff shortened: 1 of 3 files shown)


lib/new_relic/util/vendor.ex
```diff
@@ -0,0 +1,44 @@
...
+        case Jason.decode(body) do
+          {:ok, data} -> Map.take(data, @aws_vendor_data)
+          _ -> nil
+        end
+
+      _error ->
+        nil
+    end
+  rescue
+    _exception -> nil
```
mattbaker: Is it worth logging something out here? My only fear is that swallowing the error would make it harder for a user to debug something, even if it's not the agent that's the issue

binaryseed: good call

mattbaker:
nice

PR #91, opened 2019-02-01
Clear warning with left side pattern match

### LINKED ISSUE #90: @trace directives cause compilation warnings.
**Describe the bug**
When I add the `@trace` directive before a function I get an unused variable warning for `args` on the following code.

```
  @trace {:get_user_details, category: :external}
  def get_user_details(args = %{open_id: _}, headers) do
    "/api/public/v2/user/details"
    |> get(add_open_id_header(headers, args))
    |> handle_response(%{"data" => %User{}})
  end
```
the full warning is:
```
warning: variable "args" is unused

Note variables defined inside case, cond, fn, if and similar do not leak. If you want to conditionally override an existing variable "args", you will have to explicitly return the variable. For example:

    if some_condition? do
      atom = :one
    else
      atom = :two
    end

should be written as

    atom =
      if some_condition? do
        :one
      else
        :two
      end

Unused variable found at:
  lib/mana_web/services/user_service.ex:9
```

**Environment**
Erlang/OTP 21 [erts-10.0.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
Elixir 1.7.3 (compiled with Erlang/OTP 21)
new_relic_agent 1.6.2 (Hex package) (mix)
  locked at 1.6.2 (new_relic_agent) a20b514d

binaryseed:
Fixes #90 , clearing up a warning when `@trace`-ing functions that have left-side pattern matches. We unquote the args to a function to send to the reporter, and so we need to clean up pattern matching as we are outside the actual function head.

There was logic for this but it wasn't catching the condition properly

sidekick/ @mattbaker

### DIFF
--- a/lib/new_relic/tracer/macro.ex
+++ b/lib/new_relic/tracer/macro.ex
@@ -193,7 +193,7 @@ defmodule NewRelic.Tracer.Macro do
   # Search for variables on the left side of a pattern match
   # and prefix them with an underscore so they don't end up as
   # unused variables
-  def rewrite_call_term({:=, metadata, [pattern, {name, _, context} = value]})
+  def rewrite_call_term({:=, metadata, [{name, _, context} = pattern, value]})
       when is_variable(name, context) do
     ignored_pattern = Macro.postwalk(pattern, &rewrite_match_pattern/1)
     {:=, metadata, [ignored_pattern, value]}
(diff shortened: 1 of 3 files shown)


mattbaker:
Looks like this may have caused a regression around some work we did to avoid "unused variable" warnings a while back. I'll dig up the relevant PR and send it your way.

binaryseed:
Closing in favor of #92

PR #118, opened 2019-06-07
Fix: race condition when tracing async_nolink processes

zoevkay:
## Description

Discovered a race condition where, when an `async_nolink` process is started within a transaction, if it finishes after the transaction has completed, the transaction it is linked to no longer exists so the process's results will not be reported.

This PR removes the linking of an `async_nolink` process to a transaction so that it will consistently be reported to the agent. Note: when using erlang 20, the logging is async and the race condition still exists.

### DIFF
--- a/lib/new_relic/transaction/monitor.ex
+++ b/lib/new_relic/transaction/monitor.ex
@@ -54,16 +54,6 @@ defmodule NewRelic.Transaction.Monitor do
   end
 
   def handle_info(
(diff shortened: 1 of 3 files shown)


binaryseed:
Thanks @zoevkay !!

FYI, our thinking behind this is that using `async_nolink` is signaling that the work done is _not_ part of the current "Transaction", so we aren't going to try tracking it...

Issue #129, opened 2019-08-19
Use span attributes to generate External metrics


binaryseed:
When instrumenting for Distributed Tracing, we know the URL of the External call, so we can use it in favor of the standard `{module, function, arity}` to identify the external

https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic.ex#L72-L80

https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/tracer/report.ex#L100

https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/tracer/report.ex#L121-L124

`External/host/all`
`External/host/client/method`

PR #130, opened 2019-08-20
Use span attr to name external call metric

CaiqueMitsuoka:
To have the domain of the external call reported in the dashboard, we can use the span attributes  when it's a `http` span because it includes the component, method and url of the call.
But this requires a span to be reported. This PRs includes changes to use that to report the External Metric following the convention of `domain/component/method` to the NewRelic.

If the span doesn't contain the required key, it will fallback to the default and current behaviour of `module.function/arity` of the traced call.

Solves #129


binaryseed:
Hi! Sorry this one fell off my radar for a while.. I've merged it into a working branch - I need to adjust some of the details about the metrics that are reported

PR #136, opened 2019-09-13
Protect against huge arguments blowing out memory

binaryseed:
This PR is intended to protect against excessive memory usage in cases where a very large term is passed through a `@trace`d function.

This is done by `inspect`ing the term at the point of instrumentation before we store it inside a TransactionTrace or Span. `Inspect` has mechanisms for short-circuiting it's work, and we leverage those.

I tested this in our GraphQL service, which contains `Absinthe` structs that are known to be gigantic. The resulting data was 20x smaller, and avoided a full memory copy of the struct. Yay!

### DIFF
--- a/lib/new_relic/tracer/report.ex
+++ b/lib/new_relic/tracer/report.ex
@@ -11,7 +11,7 @@ defmodule NewRelic.Tracer.Report do
   @moduledoc false
 
   def call(
-        {module, function, args},
+        {module, function, arguments},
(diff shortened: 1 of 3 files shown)


mattbaker:
@binaryseed FYI `Collector.SpanEvent.Harvester` might be coming up null in that test (just glanced at travis)

mattbaker:
:sheep:

Issue #141, opened 2019-10-07
Track request queueing


sb8244:
**Is your feature request related to a problem? Please describe.**

Our app uses proxies to get through to the Phoenix server. We want to track time spent in proxy as request queueing.

**Describe the solution you'd like**

The same request queueing feature that Ruby has, based on `X-Request-Start` header

**Describe alternatives you've considered**

I have not

sb8244:
Happy to spike on this a bit, per our convo on Slack.

binaryseed:
Here's a start...

To support this, we need to report:
* A `WebFrontend/QueueTime` metric. Unit: Seconds
* A `queueDuration` attribute on the Transaction. Unit: Seconds

Inside `NewRelic.Transaction.Plug` we'd add a new `maybe_report_queuing` function which we'd call from `on_call`

We need to read the `x-request-start` header, and handle potential clock skew (ie: don't log a negative queue time)

sb8244:
Do you handle clock skew by just having a `max(timing, 0)` check? So if it did skew, it would just come across as 0?

binaryseed:
I think that makes sense, in the case of a negative queue_time, it looks like Ruby treats that as `0`.. 

* https://github.com/newrelic/rpm/blob/17734c85a9251ede14fbee517f0d264d99116c7e/lib/new_relic/agent/instrumentation/queue_time.rb#L43

Also, we need to handle `x-request-start` as either Int or Float, and Seconds / Milliseconds / Microseconds

PR #142, opened 2019-10-09
Sb8244 request queue

sb8244:
Okay, this is for #141. It's the biggest PR I've done here, so it probably has a good bit that could be changed.

Goals:

* Track `queueDuration` as seconds (fractional to the microsecond) on web transactions
* The queue start time is interpreted from one of `"x-request-start", "x-queue-start", "x-middleware-start"`, as per the Ruby implementation
* The queue start time can be microseconds, milliseconds, or seconds (fractions supported)
* Clock skew, start time in the future, is interpreted as "now"
* WebFrontend/QueueTime metric is dispatched with completed requests

The Ruby implementation served as inspiration for deriving queue start, but I didn't handle some cases like having a compound header, or multiple headers with different values.

I haven't tried it in practice, yet.

### DIFF
--- a/lib/new_relic/harvest/collector/metric_data.ex
+++ b/lib/new_relic/harvest/collector/metric_data.ex
@@ -10,49 +10,50 @@ defmodule NewRelic.Harvest.Collector.MetricData do
         total_time_s: total_time_s,
         queue_duration_s: queue_duration_s
       ),
-      do: [
-        %Metric{
-          name: :HttpDispatcher,
-          call_count: 1,
-          total_call_time: duration_s,
-          total_exclusive_time: duration_s,
-          min_call_time: duration_s,
-          max_call_time: duration_s
-        },
-        %Metric{
-          name: :WebTransaction,
-          call_count: 1,
-          total_call_time: duration_s,
-          total_exclusive_time: duration_s,
-          min_call_time: duration_s,
-          max_call_time: duration_s
-        },
-        %Metric{
-          name: join(["WebTransaction", name]),
-          call_count: 1,
-          total_call_time: duration_s,
-          total_exclusive_time: duration_s,
-          min_call_time: duration_s,
-          max_call_time: duration_s
-        },
-        %Metric{
-          name: :WebTransactionTotalTime,
-          call_count: 1,
-          total_call_time: total_time_s,
-          total_exclusive_time: total_time_s,
-          min_call_time: total_time_s,
-          max_call_time: total_time_s
-        },
-        %Metric{
-          name: join(["WebTransactionTotalTime", name]),
-          call_count: 1,
-          total_call_time: total_time_s,
-          total_exclusive_time: total_time_s,
-          min_call_time: total_time_s,
-          max_call_time: total_time_s
-        }
-      ]
-      |> queue_duration_metric(queue_duration_s)
+      do:
+        [
+          %Metric{
+            name: :HttpDispatcher,
+            call_count: 1,
+            total_call_time: duration_s,
+            total_exclusive_time: duration_s,
+            min_call_time: duration_s,
+            max_call_time: duration_s
+          },
+          %Metric{
+            name: :WebTransaction,
+            call_count: 1,
+            total_call_time: duration_s,
+            total_exclusive_time: duration_s,
+            min_call_time: duration_s,
+            max_call_time: duration_s
+          },
+          %Metric{
+            name: join(["WebTransaction", name]),
+            call_count: 1,
+            total_call_time: duration_s,
+            total_exclusive_time: duration_s,
+            min_call_time: duration_s,
+            max_call_time: duration_s
+          },
+          %Metric{
+            name: :WebTransactionTotalTime,
+            call_count: 1,
+            total_call_time: total_time_s,
+            total_exclusive_time: total_time_s,
+            min_call_time: total_time_s,
+            max_call_time: total_time_s
+          },
+          %Metric{
+            name: join(["WebTransactionTotalTime", name]),
+            call_count: 1,
+            total_call_time: total_time_s,
+            total_exclusive_time: total_time_s,
+            min_call_time: total_time_s,
+            max_call_time: total_time_s
+          }
+        ]
+        |> queue_duration_metric(queue_duration_s)
 
   def transform({:other_transaction, name}, duration_s: duration_s, total_time_s: total_time_s),
     do: [
--- a/lib/new_relic/harvest/collector/metric_data.ex
+++ b/lib/new_relic/harvest/collector/metric_data.ex
@@ -5,55 +5,49 @@ defmodule NewRelic.Harvest.Collector.MetricData do
 
   alias NewRelic.Metric
 
-  def transform({:transaction, name},
-        duration_s: duration_s,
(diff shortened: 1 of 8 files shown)


lib/new_relic/transaction/complete.ex
```diff
@@ -57,6 +58,20 @@ defmodule NewRelic.Transaction.Complete do
     })
   end
 
+  defp derive_queue_duration(%{start_time: start_time, queue_start_us: queue_start_us} = tx)
+       when not is_nil(queue_start_us) do
+    start_time_us = System.convert_time_unit(start_time, :native, :microsecond)
+    queue_duration_us = max(0, start_time_us - queue_start_us)
```
sb8244: This is one protection from skew. This could stay, or the other could stay, or both. I think this one is much more performant since it's not going to clock, so it might be a better candidate to stay.

lib/new_relic/harvest/collector/metric_data.ex
```diff
@@ -5,49 +5,55 @@ defmodule NewRelic.Harvest.Collector.MetricData do
...
+          %Metric{
+            name: join(["WebTransactionTotalTime", name]),
+            call_count: 1,
+            total_call_time: total_time_s,
+            total_exclusive_time: total_time_s,
+            min_call_time: total_time_s,
+            max_call_time: total_time_s
+          }
+        ]
+        |> queue_duration_metric(queue_duration_s)
```
binaryseed: we should report the queue metric separately from this, in this file we should just be adding a function like this:

```elixir
def transform(:queue_time, duration_s: duration_s) do
  [
    %Metric{}
  ]
end
```

lib/new_relic/transaction/complete.ex
```diff
@@ -494,7 +510,8 @@ defmodule NewRelic.Transaction.Complete do
   def report_transaction_metric(tx) do
     NewRelic.report_metric({:transaction, tx.name},
       duration_s: tx.duration_s,
-      total_time_s: tx.total_time_s
+      total_time_s: tx.total_time_s,
+      queue_duration_s: Map.get(tx, :queue_duration_s, nil)
```
binaryseed: let's have 2 calls to `report_metric` here instead of smushing them together

lib/new_relic/transaction/plug.ex
```diff
@@ -80,4 +82,30 @@ defmodule NewRelic.Transaction.Plug do
...
   end
+
+  defp maybe_report_queueing(conn) do
+    [
+      queue_start_us: extract_queue_start_us(conn)
+    ]
+    |> NewRelic.add_attributes()
+  end
+
+  @queue_duration_headers ["x-request-start", "x-queue-start", "x-middleware-start"]
```
binaryseed: I'd rather just stick with `x-request-start` than pile on a bunch of extra logic for backwards compatibility

lib/new_relic/transaction/plug.ex
```diff
@@ -80,4 +82,30 @@ defmodule NewRelic.Transaction.Plug do
...
+  defp extract_queue_start_us(%{req_headers: headers}),
+    do:
+      Enum.reduce_while(headers, nil, fn
+        {header, timestamp}, acc when header in @queue_duration_headers ->
+          case RequestQueueTime.timestamp_to_us(timestamp) do
+            {:ok, us} ->
+              {:halt, us}
+
+            {:error, reason} ->
+              Logger.debug(reason)
```
binaryseed: There's a NewRelic logger that'd be preferable, though i'm not sure it's something i'd really want to log out

lib/new_relic/transaction/event.ex
```diff
@@ -23,7 +24,8 @@ defmodule NewRelic.Transaction.Event do
         timestamp: transaction.timestamp,
         name: transaction.name,
         duration: transaction.duration,
-        type: transaction.type
+        type: transaction.type,
+        queueDuration: transaction.queue_duration
```
binaryseed: This isn't where we should be reporting the attribute, only a small set of attributes are accepted here, we can remove this from `Event`.

lib/new_relic/transaction/complete.ex
```diff
@@ -57,6 +58,20 @@ defmodule NewRelic.Transaction.Complete do
...
 
+  defp derive_queue_duration(%{start_time: start_time, queue_start_us: queue_start_us} = tx)
+       when not is_nil(queue_start_us) do
+    start_time_us = System.convert_time_unit(start_time, :native, :microsecond)
+    queue_duration_us = max(0, start_time_us - queue_start_us)
+
+    tx
+    |> Map.merge(%{
+      queue_duration_us: queue_duration_us,
+      queue_duration_s: queue_duration_us / 1_000_000
```
binaryseed: This is actually where we need to report the final `queueDuration` attribute... I don't think we need the other time unit attributes, and we should actually drop the `queue_start_us` attribute

sb8244: Oh. So these transaction attributes will be added to the final transaction automatically? The key should be queueDuration in that case?

lib/new_relic/transaction/plug.ex
sb8244: I'm happy to remove it. I put it here to mirror the Ruby agent log. I could see someone wondering "why isn't my queueing metric reporting?"

lib/new_relic/transaction/plug.ex
sb8244: Happy to make that change. I didn't realize they were legacy

lib/new_relic/harvest/collector/metric_data.ex
```diff
@@ -253,5 +253,20 @@ defmodule NewRelic.Harvest.Collector.MetricData do
       }
     ]
 
+  def transform(:queue_time, duration_s: nil), do: []
+
+  def transform(:queue_time, duration_s: duration_s) do
```
sb8244: @binaryseed I'm struggling with this metric in particular. When it was in the transaction metrics, I could easily harvest it in tests to make sure it was dispatched. I'm not sure how to test that it's dispatched now, though.

test/metric_transaction_test.exs
```diff
@@ -93,6 +94,24 @@ defmodule MetricTransactionTest do
...
+      |> Plug.Conn.put_req_header(
+        "x-request-start",
+        "t=#{:os.system_time(:microsecond) - 100_000}"
+      )
+
+    TestPlugApp.call(conn, [])
+
+    metrics = TestHelper.gather_harvest(Collector.Metric.Harvester)
+
+    assert [_, [1, time, time, time, time, 0]] =
```
sb8244: This test now fails because the metric isn't dispatched (I'm not actually sure, but would like to understand it better)

sb8244: Nevermind, it was operator error and works like I expected it to.

lib/new_relic/harvest/collector/metric_data.ex
sb8244: Cancel.

binaryseed:
Thanks for contributing! I'm going to follow up with a few commits before pulling this into `master` and cutting a release...

PR #145, opened 2019-10-14
Do not sample when newrelic is disabled

### LINKED ISSUE #144: Sampling enabled when `NewRelic.Config.enabled? == false`
**Describe the bug**

When running in a test environment we have disabled NewRelic agent through the configuration, however, sampling still occurs. This presents a problem if the process being sampled has already exited, as `Process.info/2` called in  [sampler/process.ex#L75](https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/sampler/process.ex#L75) returns `nil`. This results in an argument error as seen below: 

```
2019-08-14 13:16:08.501 [error] GenServer NewRelic.Sampler.Process terminating
** (ArithmeticError) bad argument in arithmetic expression
    (new_relic_agent) lib/new_relic/sampler/process.ex:79: NewRelic.Sampler.Process.take_sample/1
    (new_relic_agent) lib/new_relic/sampler/process.ex:62: NewRelic.Sampler.Process.store_pid/3
    (new_relic_agent) lib/new_relic/sampler/process.ex:24: NewRelic.Sampler.Process.handle_cast/2
    (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:711: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: {:"$gen_cast", {:sample_process, #PID<0.6739.0>}}
State: %{pids: %{#PID<0.1314.0> => true, #PID<0.1315.0> => true, ..., reductions: 46}}}
```

**Environment**
```
Erlang/OTP 21 [erts-10.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.9.1 (compiled with Erlang/OTP 21)
```

```
new_relic_agent 1.9.11
```

> Note: Although we are a few versions behind, the code path for this still exists in `master`.

erikreedstrom:
This PR allows sampling to be disabled when `NewRelic.Config.enabled?` returns `false`.

> Note: To ensure tests complete, a test helper has been added in the style of existing helpers.

**Alternate Solution**
The underlying issue described in #144 is an argument error when trying to calculate memory. 
Another approach considered to resolve was handling the `nil` case, however, it seemed odd that we would be sampling at all given nothing is reported.

Resolves #144

### DIFF
--- a/lib/new_relic/sampler/process.ex
+++ b/lib/new_relic/sampler/process.ex
@@ -18,7 +18,11 @@ defmodule NewRelic.Sampler.Process do
     {:ok, %{pids: %{}, previous: %{}}}
   end
 
-  def sample_process, do: GenServer.cast(__MODULE__, {:sample_process, self()})
+  def sample_process do
+    if NewRelic.Config.enabled?(),
+      do: GenServer.cast(__MODULE__, {:sample_process, self()}),
+      else: :ok
+  end
 
   def handle_cast({:sample_process, pid}, state) do
     state = store_pid(state.pids[pid], state, pid)
--- a/test/sampler_test.exs
+++ b/test/sampler_test.exs
@@ -48,8 +48,9 @@ defmodule SamplerTest do
 
   test "Process Sampler" do
     TestHelper.restart_harvest_cycle(Collector.CustomEvent.HarvestCycle)
-    TestProcess.start_link()
+    {:ok, process} = TestProcess.start_link()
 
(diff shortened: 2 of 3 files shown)


binaryseed:
Hello! Thanks for the PR. There's quite a few things that we can avoid doing when the agent is disabled, so I opened up #146 to address them more generally...

binaryseed:
Closing in favor of the other PR

PR #154, opened 2019-10-24
Update README.md: "start new_relic application"

amcvitty:
While installing, I didn't immediately realise it was necessary to start the new_relic application, and my instrumentation was failing.

I believe this is the right thing to do. It certainly got my installation working! If there is a different way to do it, then I suggest that should be documented instead. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,22 @@ defp deps do
 end
 ```
 
+Ensure new_relic is started before your application:
+
+```elixir
+def application do
+  [
+    ...,
+    applications: [
+      :new_relic_agent,
+      ...
+      :myapp,
+      ...
+   ]
+  ]
+end
+```
+
 ## Configuration
 
 You need to set a few required configuration keys so we can authenticate properly.


binaryseed:
Hi there! What `elixir` version are you using? Applications should be auto-included in recent Elixir versions.

Perhaps you should remove the `applications` key altogether, and use `extra_applications`...

https://www.amberbit.com/blog/2017/9/22/elixir-applications-vs-extra_applications-guide/

amcvitty:
Well today I learned! Thanks for that. I think we originally generated our mix.exs with phoenix.new back in phoenix 1.2 and elixir 1.2.

I guess if no-one else is complaining, then there's no need to make a change. That said, if you feel like it's worth adding a comment saying "this is an application, not just a library, so if you have an explicit list of applications, please add it", then that would have helped me, so I'd support that!

binaryseed:
I don't think we quite need it in the readme, that's pretty general Elixir stuff and not so related to this particular package.. Thanks!

PR #161, opened 2020-01-04
Ecto instrumentation via `telemetry`

### LINKED ISSUE #122: Report & trace database operations
#### A) How the `@trace` works for `:datastore` events

We're attempting to adopt New Relic as Elixir monitoring tool and have stumbled upon some limitations of `new_relic_agent`. We've picked that `NewRelic.Tracer` provides the `@trace {:op, {category: :datastore}}` decorator form as the only available means for tracing database operations but NR reports generated by this API are lacking in places that are taken for granted e.g. in Ruby library. This includes:

- **Databases >> [pick operation] >> Time consumption by caller** - we should be able to see a chart with caller transactions and an ability to click any of them to inspect the situation from its end

- **Databases >> [pick operation] >> Slow queries** - we should be able to see slowest SQL statements and click on them to see the full SQL and stack trace of that specific call

- **Transactions >> [pick transaction] >> Breakdown table** - database operations should be tagged with category "Database" and the name in "Segment" column should be a link to relevant operation in Databases section

- **Transactions >> [pick transaction] >> Transaction traces >> [pick trace] >> Database queries** - we should be able to see a list of all database operations involved in specific transactions grouped by SQL with a total duration and number of calls (useful for picking N+1 issues)

- **Transactions >> [pick transaction] >> Transaction traces >> [pick trace] >> Trace details** - database operations should have the database icon which expands upon clicking into specific SQL query details and stack trace like in Databases section

This basically means that we're missing on ability to cross-reference transactions and database operations back and forth, an ability that's SUPER useful for database-centric applications such as ours, but it's the case of most Phoenix projects.

#### B) The lack of "post-factum" API alternative to `@trace`

In addition, with the `NewRelic.Tracer` as the only API for reporting spans with `"datastore"` category, we're missing on ability to report spans post-factum e.g. when reports come through telemetry only after the operation ends - such as the Ecto case, also important for most Phoenix users. 

I can only guess that this is why the effort started in https://github.com/binaryseed/new_relic_ecto/pull/4 was not finished, but also wasn't taken over by the community - because the base library seems to lack required API. As established in #53, such API should of course be database- and library-agnostic and is not expected to be automatic - it should merely lay out the foundation for `new_relic_ecto` and similar.

I believe we're talking here about one of biggest blockers for pushing forward adoption of New Relic in Elixir/Phoenix projects for many companies such as ours which successfully use & love NR with many stacks & for years but got stuck on Elixir app database monitoring.

## Solution proposal

It'd be great to see API similar to following:

```elixir
NewRelic.report_datastore_event(
  datastore, # module or arbitrary string
  resource, # it may be table name, schema name or misc DB resource name string
  operation, # insert / update / delete / select... - arbitrary string 
  query, # sql statement or equivalent for any (e.g nosql) data store
  duration # in nanoseconds as returned from System.monotonic_time() that the agent uses
)
```

With sample call as follows:

```elixir
NewRelic.report_datastore_event(
  MyApp.Repo,
  "users",
  "insert",
  "SELECT count(id) FROM users",
  5_000
)
```

This function would report all kinds of transaction spans, distributed traces and aggregates such as `NewRelic.Tracer.Report.call/5` does, but fill in the aforementioned gaps of that function, i.e.

- being callable post-factum
- making database events appear as such in transactions
- reversely - linking transactions to database operations
- including stacktrace and statement string in database operations
- naming database operation with actual datastore name instead of `"Database"`

---

What may be helpful for resolving this issue and useful for those looking for instant fill-in for Ecto instrumentation is our current solution available in https://github.com/surgeventures/new_relic_addons that tries to get the most out of current agent API in order to bring in Ecto instrumentation and some handy function decorators similar to `@trace`.

I'm not familiar with NR APIs and can't work around this issue myself, but please let me know if I can be of any assistance here. Thanks!

binaryseed:
This PR:
* Adds Ecto auto-instrumentation based on `telemetry`
* Creates `examples` directory for example apps that are fully instrumented & tested alongside agent builds

------

closes #122


binaryseed:
TODO: 
* [ ] Adopt an approach of iterating through all modules to determine which are `Ecto.Repo`s instead of tracing

mlepinay:
Any update on this?
Would love to have access to this.

binaryseed:
I released this as `1.15.0-rc.1` since it's adding a big new feature, feel free to try it out and I'll release `1.15.0` soon

PR #173, opened 2020-02-04
Handle Postgrex error results

tpitale:
When there is an error (e.g., a failed constraint) Ecto returns a result with an error tuple, instead of ok. This would raise as an unsupported Ecto adapter. Instead, we should ignore those results (at least until we can do something with them).


tpitale:
@binaryseed Would you like me to write a test for this?

binaryseed:
I think adding a query inside the example app that generates an error would be good, especially if it's placed before the other queries - that'll validate that the handler wasn't removed...

* https://github.com/newrelic/elixir_agent/blob/master/examples/apps/ecto_example/lib/ecto_example/router.ex#L9

Issue #181, opened 2020-02-26
Application fails to start if can't connect to collection host


bgroupe:
**Describe the bug**
Similar to/Regression of #131
When starting an application during an APM outage at the time of this writing 2/25/2020 application fails to start presumably if host ping times out. Full exception:

~~~
00:16:11.478 [info] Application new_relic_agent exited: NewRelic.Application.start(:normal, []) returned an error: shutdown: failed to start child: NewRelic.EnabledSupervisor
    ** (EXIT) exited in: GenServer.call(NewRelic.Harvest.Collector.AgentRun, :ping, 5000)
        ** (EXIT) time out
[os_mon] memory supervisor port (memsup): Erlang has closed
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
{"Kernel pid terminated",application_controller,"{application_start_failure,new_relic_agent,{{shutdown,{failed_to_start_child,'Elixir.NewRelic.EnabledSupervisor',{'EXIT',{timeout,{'Elixir.GenServer',call,['Elixir.NewRelic.Harvest.Collector.AgentRun',ping,5000]}}}}},{'Elixir.NewRelic.Application',start,[normal,[]]}}}"}
Kernel pid terminated (application_controller) ({application_start_failure,new_relic_agent,{{shutdown,{failed_to_start_child,'Elixir.NewRelic.EnabledSupervisor',{'EXIT',{timeout,{'Elixir.GenServer',ca
Crash dump is being written to: erl_crash.dump...done
~~~

This is problematic for running applications with self-healing orchestrators ie Kubernetes. 

**Environment**
* Elixir Version: 1.10.0-otp-22
* Erlang Version: 22.2.4
* Agent version: 1.16.4

binaryseed:
Thanks for this report! I'll take a closer look

damonvjanis:
I'm seeing this same issue, but in local development and I can't actually use any version above 1.14.0 or I get that error on startup.

Here are my replication steps:

1. `mix new my_app`
2. Add `{:new_relic_agent, "~> 1.0"} to `mix.exs`
3. Add `config :new_relic_agent, app_name: "my_app", license_key: "<valid_license_key>"` to `config.exs`
4. Run `iex -S mix`, and it hangs starting up and then crashes with the error.

I've had a co-worker run the same on his machine and there's no crash (consistently), and on mine it crashes every time, so this is potentially difficult to replicate.

[...]

PR #184, opened 2020-02-27
Boot blocking prevention

binaryseed:
This PR addresses #181 by making sure that we start the `EnabledSupervisor` only when we get confirmation that the agent is fully connected to New Relic

sidekick/ @mattbaker @tpitale

### DIFF
--- a/lib/new_relic/application.ex
+++ b/lib/new_relic/application.ex
@@ -11,7 +11,7 @@ defmodule NewRelic.Application do
     children = [
       worker(NewRelic.Logger, []),
       supervisor(NewRelic.AlwaysOnSupervisor, []),
-      supervisor(NewRelic.EnabledSupervisor, []),
+      worker(NewRelic.EnabledSupervisorManager, []),
       supervisor(NewRelic.TelemetrySupervisor, []),
       worker(NewRelic.GracefulShutdown, [], shutdown: 30_000)
     ]
--- a/lib/new_relic/enabled_supervisor.ex
+++ b/lib/new_relic/enabled_supervisor.ex
@@ -7,7 +7,6 @@ defmodule NewRelic.EnabledSupervisor do
   @moduledoc false
 
   def start_link() do
-    NewRelic.Harvest.Collector.AgentRun.ensure_init()
     start_link(enabled: NewRelic.Config.enabled?())
   end
 
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -1,7 +1,3 @@
-unless System.get_env("NR_INT_TEST") do
-  {:ok, _} = NewRelic.EnabledSupervisor.start_link(enabled: true)
-end
-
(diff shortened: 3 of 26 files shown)


lib/new_relic/enabled_supervisor.ex
```diff
@@ -7,7 +7,6 @@ defmodule NewRelic.EnabledSupervisor do
   @moduledoc false
 
   def start_link() do
-    NewRelic.Harvest.Collector.AgentRun.ensure_init()
```
binaryseed: Instead of blocking here, we rely on getting a message back inside `EnabledSupervisorManager` before starting this supervisor

test/test_helper.exs
```diff
@@ -1,7 +1,3 @@
-unless System.get_env("NR_INT_TEST") do
-  {:ok, _} = NewRelic.EnabledSupervisor.start_link(enabled: true)
```
binaryseed: We start the supervisor individually in each test that needs it instead of globally, so that we can test it starting explicitly

mattbaker:
Intuitive, makes sense.

Weird unrelated test failure

```
  1) test Beam stats Sampler (SamplerTest)
     test/sampler_test.exs:32
     Assertion with > failed
     code:  assert cpu > 0
     left:  0.0
     right: 0
     stacktrace:
       test/sampler_test.exs:53: (test)
```

lib/new_relic/application.ex
```diff
@@ -11,7 +11,7 @@ defmodule NewRelic.Application do
     children = [
       worker(NewRelic.Logger, []),
       supervisor(NewRelic.AlwaysOnSupervisor, []),
-      supervisor(NewRelic.EnabledSupervisor, []),
+      worker(NewRelic.EnabledSupervisorManager, []),
```
tpitale: This might be too much, but should this be a DynamicSupervisor, and the EnabledSupervisor would then be a child of it when started? Or do we feel like `start_link()` is enough?

tpitale: Without more than one child, it probably doesn't make sense. Answered my own question. Ignore me 😝

binaryseed: That's an interesting thought, I went this direction so that the manager process could send a message and get one back from `AgentRun` to signal when to start the `EnabledSupervisor`. The original problem was that there was a blocking call in `start_link` of `EnabledSupervisor`, so this made it async.

It would be nice if it was all supervisors w/o a `GenServer` stuck in between, perhaps `AgentRun` could trigger the `EnabledSupervisor` to start as part of it's connect process

binaryseed: Tried it out in https://github.com/newrelic/elixir_agent/pull/187 , I like that it uses standard supervisors the whole way down

binaryseed:
Going with #187 instead

PR #186, opened 2020-02-27
Enable ignoring arguments in traces

### LINKED ISSUE #157: Prevent logging of @trace args
**Is your feature request related to a problem? Please describe.**
Traced functions are so fast to use and work great, but they include the `args` of the function, and that contains a private API key in my case.

**Describe the solution you'd like**
I'd like all of the same tracing, but the args marked as `[]` or `[:redacted]`

**Describe alternatives you've considered**
I am thinking about finding a way to structure the function without accepting the input. Unfortunately I don't see a way yet. I am also considering replicating what `@trace` does, but that ends up being fairly complex.

binaryseed:
This PR adds a feature toggle around function argument tracing so that it can be disabled if sensitive information is in the arguments. Also if lots of tracing is happening, this reduces memory footprint

closes #157

### DIFF
--- a/lib/new_relic/config.ex
+++ b/lib/new_relic/config.ex
@@ -97,6 +97,8 @@ defmodule NewRelic.Config do
     * Controls collection of SQL query strings
   * `:ecto_instrumentation_enabled` (default `true`)
     * Controls all Ecto instrumentation
+  * `function_argument_collection_enabled` (default `true`)
+    * Controls collection of traced function arguments
   """
   def feature?(:error_collector) do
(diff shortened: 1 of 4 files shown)


mattbaker:
It'd be sweet if this could be per-function somehow too, like if I could annotate just `my_secret/2` with an instruction to not record args

binaryseed:
Yeah, added an option for individual function traces, that'll be nice

sb8244:
Thanks!

binaryseed:
This will make it into the upcoming `1.17` release!

PR #187, opened 2020-02-27
Boot block prevention take 2

binaryseed:
Alternative to #184 that uses a `DynamicSupervisor` to let us complete the boot cycle after we know we've connected inside `AgentRun`

This removes the blocking call that we previously relied on to ensure boot order.

fixes https://github.com/newrelic/elixir_agent/pull/184

@mattbaker @tpitale

### DIFF
--- a/lib/new_relic/harvest/collector/agent_run.ex
+++ b/lib/new_relic/harvest/collector/agent_run.ex
@@ -30,8 +30,6 @@ defmodule NewRelic.Harvest.Collector.AgentRun do
 
   def reconnect, do: send(__MODULE__, :reconnect)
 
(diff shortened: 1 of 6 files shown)


lib/new_relic/harvest/collector/agent_run.ex
```diff
@@ -45,7 +43,16 @@ defmodule NewRelic.Harvest.Collector.AgentRun do
 
   def handle_continue(:connect, _state) do
     status = connect()
-    {:noreply, %{status: status}}
+    {:noreply, %{status: status}, {:continue, :complete_boot}}
```
tpitale: 👍

tpitale:
I like this one better, as it turns out.

PR #197, opened 2020-03-03
Use built-in SSL hostname check

binaryseed:
OTP 21+ (which we require) has a built-in SSL hostname verification function that we can use instead of relying on a dependency...

Described in the Erlang Elixir Foundation Security guide:
* https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl


mattbaker:
Looks reasonable

PR #205, opened 2020-03-13
Coercion of custom attributes

binaryseed:
This PR handles user-supplied attributes of types that aren't normally JSON serializable.

A set of types we can easily `inspect` or `to_string` to get a meaningful value. Another set we reject since we can't serialize it to JSON in a meaningful way - like tuples / functions.

### DIFF
--- a/lib/new_relic/util.ex
+++ b/lib/new_relic/util.ex
@@ -41,6 +41,10 @@ defmodule NewRelic.Util do
     |> Enum.concat([{"#{key}.length", length(list_value)}])
   end
 
+  def deep_flatten({key, %{__struct__: _} = value}) do
+    [{key, value}]
+  end
+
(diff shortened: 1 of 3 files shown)


lib/new_relic/util.ex
```diff
@@ -52,6 +56,36 @@ defmodule NewRelic.Util do
...
+          false -> bad_value(key, value)
+        end
+
+      {key, value} when is_reference(value) when is_pid(value) when is_port(value) ->
+        {key, inspect(value)}
+
+      {key, value} when is_atom(value) ->
+        {key, to_string(value)}
+
+      {key, %struct{} = value} when struct in [Date, DateTime, Time, NaiveDateTime] ->
```
binaryseed: stringerize these to ISO-8601 instead of inspect

Issue #204, opened 2020-03-11
Report Docker container ID


binaryseed:
Reporting the Docker Container ID helps wire up some UIs, including services running in Kubernetes

binaryseed:
Will be included in 1.18

PR #207, opened 2020-03-21
Add docker id to the utilization vendors

alejandrodnm:
If the agent detects that the app is running inside a container the container ID is added to the utilization map as:

%{docker: %{id: "<CONTAINER_ID>"}}

The conatainer ID is taken from the file "/proc/self/cgroup". It's taken from the first cgroup in the file that meets the following requirements: 

- The definition must contain 3 parts separated by a colon.
- It must specify the cpu subsystem.
- The contianer ID is the last item in the cgroup path and it should be a 64-digit hex string. 

Fixes https://github.com/newrelic/elixir_agent/issues/204

### DIFF
--- a/lib/new_relic/util/vendor.ex
+++ b/lib/new_relic/util/vendor.ex
@@ -1,13 +1,81 @@
 defmodule NewRelic.Util.Vendor do
   @moduledoc false
 
+  @cgroup_file "/proc/self/cgroup"
+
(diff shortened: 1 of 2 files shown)


binaryseed:
Thanks, this is super helpful! I hope to get this pulled in next week (though things in the world are a bit crazy right now)

binaryseed:
Thanks, I'm going to merge into a feature branch for a few tweaks!

PR #210, opened 2020-03-31
Redix instrumentation

binaryseed:
This PR adds automatic `redix` instrumentation!

This instrumentation is on by default, and can be controlled with:
* `redix_instrumentation_enabled: false` to turn the whole thing off
* `db_query_collection_enabled: false` to opt-out of query string collection for security


lib/new_relic/telemetry/redix.ex
```diff
@@ -0,0 +1,185 @@
...
+      update_in(
+        config.connections,
+        &Map.put(
+          &1,
+          meta.connection,
+          %{address: meta.address, host: host, port: port}
+        )
+      )
+
+    :telemetry.detach(config.handler_id)
```
binaryseed, line 48: Attach with additional handlers instead of re-attaching. Each handler will contain just it's config.

lib/new_relic/telemetry/redix.ex
```diff
@@ -0,0 +1,170 @@
...
+        %{
+          component: datastore,
+          "span.kind": :client,
+          "db.statement": query,
+          "peer.address": address,
+          "peer.hostname": hostname,
+          "db.operation": operation,
+          "redix.connection": inspect(connection)
+        }
+        |> maybe_add("redix.error", meta[:error])
```
binaryseed, line 127: * create a test situation that exercises this, I think it's changed

https://github.com/whatyouhide/redix/commit/3cd9beaa9a168fa3412d0d4b505b55ca76acc8cb#diff-d0c8798ead6df7f075d966a50c6d0fbbR110

binaryseed: https://github.com/whatyouhide/redix/commit/3cd9beaa9a168fa3412d0d4b505b55ca76acc8cb#diff-f27faf01f440bea783cd29f2ccf82345R640-R641

PR #211, opened 2020-04-13
Tesla instrumentation

andrewhr:
Based on recent changes around [Tesla telemetry events][1]. This patch is based on the work to get [Redix][2] and follow kind of same patterns. A bit of metrics used are copied from our existing "external" helpers, while metric name pattern (External/example.com/Tesla/GET) follows the conventions found on [Ruby agent implementation][3].

Since a lot of assumptions were take regarding how those metrics would look like, and it is not clear to me the patterns and naming conventions for NewRelic, I decided to open this PR for preliminary inspection.

I still need to work around some tests and example applications (like I found for other telemetry-based implementation), but hope this is a good start towards the [Telemetry adoption goal][4].

[1]: https://github.com/teamon/tesla/blob/02e42e69b7dea393e804b24e05824b17493f71e2/lib/tesla/middleware/telemetry.ex#L27
[2]: https://github.com/newrelic/elixir_agent/pull/210/commits/fa15abc3550d51e6366d0b856dafc6f50a5adead#diff-af47d0f3bef9fb5b4885207e21512992R1
[3]: https://github.com/newrelic/rpm/blob/f42f233c02c3ad85ed4a8b41ea9bf2a664103d26/lib/new_relic/agent/transaction/external_request_segment.rb#L40
[4]: https://github.com/newrelic/elixir_agent/projects/1


binaryseed:
Hi, thanks for this PR! I think we'll add this but what I'm unsure about yet is how to get support for Distributed Tracing alongside metric & span capture. 

There's a ticket where I've been proposing some ideas to do this via Telemetry:

https://github.com/beam-telemetry/telemetry/issues/48

I don't think this should necessarily prevent us from adding this instrumentation, but it does create a situation where externals will be captured via telemetry auto-instrumentation, but DT doesn't work.

andrewhr:
You have a good point.

We do have a thin Tesla.Middleware that adds `NewRelic.distributed_trace_headers(:http)` to the request, but is a manual step required for each client you want distributed tracing to work.

binaryseed:
Thanks for this. I'm going to close for now, but will re-visit in the future.

https://github.com/newrelic/elixir_agent/projects/1#card-42316968

PR #215, opened 2020-04-28
Connect Browser Distributed Traces

binaryseed:
The New Relic Browser agent now supports DT, and there are some incorrect assumptions in the agent about which attributes will be present in the payload.

This PR fixes those assumptions to match what actually happens - there's no "transaction" (`tx`) attribute.

A few more fixes are included:
* Properly encode two DT payload values as strings
* Match the latest expected DT behavior & always send the Span ID
* Match expected nil attribute value behavior - don't coerce nil values, instead exclude them

### DIFF
--- a/lib/new_relic/distributed_trace.ex
+++ b/lib/new_relic/distributed_trace.ex
@@ -85,7 +85,7 @@ defmodule NewRelic.DistributedTrace do
 
   def maybe_generate_sampling(context), do: context
 
-  def maybe_generate_trace_id(%Context{parent_id: nil} = context) do
+  def maybe_generate_trace_id(%Context{trace_id: nil} = context) do
     %{context | trace_id: generate_guid(16)}
   end
 
@@ -112,7 +112,10 @@ defmodule NewRelic.DistributedTrace do
   end
 
   def report_attributes(
-        %Context{parent_id: nil} = context,
+        %Context{
+          parent_id: nil,
+          span_guid: nil
+        } = context,
         transport_type: _type
       ) do
     [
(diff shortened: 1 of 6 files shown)


lib/new_relic/distributed_trace.ex
```diff
@@ -112,7 +112,10 @@ defmodule NewRelic.DistributedTrace do
   end
 
   def report_attributes(
-        %Context{parent_id: nil} = context,
+        %Context{
+          parent_id: nil,
+          span_guid: nil
+        } = context,
```
binaryseed, line 118: Only when there's no "id" AND no "tx" on the incoming payload do we consider there to be no parent, otherwise we do and the `report_attributes` function head below is executed

justinfoote: So, if either of them is nil, the other `report_attributes` function is used, and either `parentId` or `parentSpanId` is set to nil.  And then, somewhere else, the nil attributes will be excluded from the event sent to the back end, right?

lib/new_relic/distributed_trace/new_relic_context.ex
```diff
@@ -77,20 +77,17 @@ defmodule NewRelic.DistributedTrace.NewRelicContext do
           "ap" => context.app_id |> to_string,
           "tx" => context.guid,
           "tr" => context.trace_id,
+          "id" => context.span_guid,
```
justinfoote, line 80: 👍

lib/new_relic/util.ex
```diff
@@ -57,27 +57,30 @@ defmodule NewRelic.Util do
   end
 
   def coerce_attributes(attrs) do
-    Enum.map(attrs, fn
+    Enum.flat_map(attrs, fn
+      {_key, nil} ->
+        []
```
justinfoote, line 62: ❤️

lib/new_relic/distributed_trace.ex
binaryseed: Right, if _both_ are nil, this is the function head that gets picked, otherwise the one below. Then later on all `nil` values get dropped like the spec says they should have been all along :)

PR #217, opened 2020-05-03
Add reporter fail to public api

prabello:
Delegate the transaction fail on the public API and fix the reference on the doc.


binaryseed:
Thanks for this! The transaction should have been properly marked as an error actually, so I prepared this PR to fix that issue: #220

PR #218, opened 2020-05-03
Add Broadway instrumentation

andrewhr:
Add instrumentation based on Telemetry events. [Broadway translates itself to a topology of GenStage processes][1], and that's what exposed through its [telemetry events][2].

The strategy taken was then to use each GenStage start/stop event to create a NewRelic transaction boundary. This does create a bit of indirection as Broadway callback module talks in terms of `handle_message/3` and `handle_batch/4`, but on the good side, we get closer to what "is happening", get a better mapping of telemetry events to transactions. Also, the docs are extensively describing those stages.

To alleviate the case of Processors - the GenStage works with batches, but Broadway callback calls each message individually - the instrumentation here creates separate segments/spans for each processed message.

The Batcher process was initially added here, but removed before publishing this commit because 1-) it's usually quite fast as its logic is simple 2-) there is not much the user can do about it, as there is a way to add custom code yet.

### Pending

- ~[ ] Add failures just to each message span instead of the whole transaction (I don't know if that's possible)~ Doesn't apply anymore, each `handle_message/3` will be wrapped on their own Transaction, so errors will be applied on that transaction.
- ~[ ] Make message spans "parents": my tests run Ecto on each handle_message, but they become siblings instead of children on DT UI. I experiment a bit with `:nr_current_span` stacks but didn't work well~ Doesn't apply anymore, each `handle_message/3` will be wrapped on their own Transaction
- [x] Add tests!

[1]: https://hexdocs.pm/broadway/0.6.0/architecture.html
[2]: https://hexdocs.pm/broadway/0.6.0/Broadway.html#module-telemetry


lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,190 @@
...
+
+    Process.flag(:trap_exit, true)
+    {:ok, config}
+  end
+
+  def terminate(_reason, %{handler_id: handler_id}) do
+    :telemetry.detach(handler_id)
+  end
+
+  def handle_event(@processor_start, _measurements, metadata, _config) do
```
binaryseed, line 53: How long does the "processor" live? If it doesn't have a short lifecycle (from seconds to a few mins), we can't use it as the Transaction, as it'd just build up memory and risk OOM kills

lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,190 @@
...
+  def handle_event(@processor_stop, _measurements, metadata, _config) do
+    NewRelic.add_attributes(processor_stop_attributes(metadata))
+    Transaction.stop_transaction()
+  end
+
+  def handle_event(@message_stop, measurements, metadata, _config) do
+    track_message_segment(measurements, metadata)
+  end
+
+  def handle_event(@message_fail, measurements, metadata, _config) do
```
binaryseed, line 70: If there was already an exception raised & caught, it's possible this error state is already handled. We'll need to verify this with tests of that condition

lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,190 @@
...
+        error \\ nil
+      ) do
+    end_time_ms = System.system_time(:millisecond)
+    duration_ms = System.convert_time_unit(duration, :native, :millisecond)
+    duration_s = duration_ms / 1000
+    start_time_ms = end_time_ms - duration_ms
+
+    pid = inspect(self())
+    id = {:broadway_message, make_ref()}
+    parent_id = Process.get(:nr_current_span) || :root
```
binaryseed, line 132: I'm not 100% sure we want to model this with a Transaction for the processor & segments for messages, but that aside, I think I'll want to make a helper function for fetching the "parent" / "id" stuff instead of duplicating this code in a few places, since it's not super easy...

lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,190 @@
...
+      "broadway.stage": :processor,
+      "broadway.processor_message_count": length(messages)
+    ]
+  end
+
+  defp processor_stop_attributes(metadata) do
+    info = Process.info(self(), [:memory, :reductions])
+
+    [
+      "broadway.processor_successful_to_ack_count": length(metadata.successful_messages_to_ack),
```
binaryseed, line 111: I'm thinking the attribute keys should more closely mirror the names they have in the broadway metadata...

Like `broadway.processor.successful_messages_to_ack` for this one for example

lib/new_relic/telemetry/broadway.ex
andrewhr: Good point, I'm not sure either. On my first tests looked a good idea, but with some more real word testing I'm seeing really huge transaction traces that may not be ideal to follow.

lib/new_relic/telemetry/broadway.ex
andrewhr: Here goes a problem: Broadway captures all the errors and logs then, but it is not clear to me if they end on `error_logger`. Maybe @msaraiva may help, or we can open an issue on Broadway repo?

lib/new_relic/telemetry/broadway.ex
andrewhr: The processor process can leave as long the app lives, but this transaction runs for each "batch". I'm running this fork on my app for testing, there are ~10 minute long transactions. If you wish, I can send you an email with a link for you to access this app.

lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,190 @@
...
+    Process.flag(:trap_exit, true)
+    {:ok, config}
+  end
+
+  def terminate(_reason, %{handler_id: handler_id}) do
+    :telemetry.detach(handler_id)
+  end
+
+  def handle_event(@processor_start, _measurements, metadata, _config) do
+    Transaction.start_transaction("Broadway", processor_name(metadata))
```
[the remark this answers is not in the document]
andrewhr: Fixed! 👍

lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,190 @@
...
+    NewRelic.add_attributes(processor_start_attributes(metadata))
+
+    DistributedTrace.generate_new_context()
```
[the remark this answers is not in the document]
andrewhr: Removed 👍

lib/new_relic/telemetry/broadway.ex
binaryseed: Yeah, if broadway rescues them, they probably don't make it too logger at all. We'll want to exercise this in a simple test though

msaraiva: Hi folks!

All errors, unless explicitly wrapped in a `try/catch` block, should be captured by Broadway and logged using `Logger.error/2`. In case there's any scenario you have which is not covered and you need any additional feature, please feel free to open an issue at https://github.com/dashbitco/broadway for discussion.

Cheers.

lib/new_relic/telemetry/broadway.ex
andrewhr: Pushed a change here to make a single Transaction around each `handle_message/3`, so this code is note relevant anymore.

lib/new_relic/telemetry/broadway.ex
andrewhr: Pushed a change here to make a single Transaction around each `handle_message/3`, so this particular piece of code is note relevant anymore.

On `consumer` I had a similar pattern, and applied this changes as requested 👍

lib/new_relic/telemetry/broadway.ex
andrewhr: Change strategy, single Transaction per `handle_message/3`. This issue can appear on Consumer/`handle_batch/4` but I don't think there is much to do there - that's what it does, process a batch of messages.

I'll mark this as resolved, but please let me know if you have more thoughts around this strategy

lib/new_relic/telemetry/broadway.ex
```diff
@@ -0,0 +1,216 @@
+defmodule NewRelic.Telemetry.Broadway do
+  @moduledoc """
```
andrewhr, line 2: @binaryseed I've followed documentation for Plug instrumentation, actually even copied a bulk of it! Add some description and examples how this will map on NewRelic.

I'm not a native speaker, so extra dose of review here is more than welcome 🤓

andrewhr:
Updated against latest master (1.19.2). Can you take a look how it's looks like and what needs improvement @binaryseed? Appreciate! 🙇

PR #219, opened 2020-05-06
Instrument Plug using telemetry

binaryseed:
WIP - This PR updates the `Plug` instrumentation to leverage the `telemetry` events emitted by `plug_cowboy` and `plug`

TODO:
* [x] Use duration from plug instrumentation
* [x] Depend on next version of plug
* [ ] Make sure we handle error responses sent by cowboy itself
  * https://github.com/elixir-plug/plug_cowboy/pull/34
* [x] Make sure phoenix `plug_adapter` events don't collide
  * Phoenix specific instrumentation will come later


lib/new_relic/telemetry/plug.ex
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  defp plug_name(conn),
+    do:
+      "/Plug/#{conn.method}/#{match_path(conn)}"
+      |> String.replace("/*glob", "")
+      |> String.replace("/*_path", "")
+
+  defp match_path(conn) do
+    case conn.private[:plug_route] do
```
binaryseed, line 157: For error case, this won't be set since the error is a non-local return

Need to test & validate we get something useful for the name

lib/new_relic/telemetry/plug.ex
```diff
@@ -0,0 +1,184 @@
...
+    MyTaskSupervisor,
+    fn -> do_work() end,
+    new_relic: :no_track
+  )
+  ```
+  """
+
+  alias NewRelic.{Transaction, DistributedTrace, Util}
+
+  # TODO: Should Telemetry GenServers only be started if NewRelic.Config.enabled?()
```
andrewhr, line 29: What do you think about leaving this logic only to `NewRelic.Telemetry.Supervisor`?

This should cover the requirement and avoid duplication - I can see this pattern becoming very repetitive as more telemetry instrumentation get added.

binaryseed: I'll probably just move `TelemetrySupervisor` under `EnabledSupervisor` which already does this logic...

lib/new_relic/util/http.ex
```diff
@@ -16,6 +16,14 @@ defmodule NewRelic.Util.HTTP do
   def post(url, body, headers),
     do: post(url, Jason.encode!(body), headers)
 
+  def get(url, headers) do
+    headers = Enum.map(headers, fn {k, v} -> {'#{k}', '#{v}'} end)
+
+    with {:ok, {{_, status, _}, _, body}} <- :httpc.request(:get, {'#{url}', headers}, [], []) do
```
tpitale, line 22: With no other match here, would this be fine as a simple assignment, without using `with`?

binaryseed: I think that'd change the behavior - it'd fail the match instead of returning the unmatched value directly...

lib/new_relic/util.ex
```diff
@@ -20,6 +20,10 @@ defmodule NewRelic.Util do
     end
   end
 
+  def get_req_header(%{req_headers: headers}, key) when is_binary(key) do
+    for {^key, value} <- headers, do: value
```
tpitale, line 24: I've not seen this form of key access before. Is it so it returns a list?

tpitale: Is this the same as doing: `headers[key] |> List.wrap()`?

tpitale: I just see a lot of `Util.get_req_header(conn, "referer") |> List.first()`, usage.

tpitale: Or maybe there can be N values with the same key …?

lib/new_relic/util/http.ex
tpitale: Ah, I didn't realize that `with` would do that, or that the unmatched value should be returned. Cool!

binaryseed:
FYI, the Plug `telemetry` instrumentation is getting reworked since it had un-reliable behavior due to exceptions being used as flow control.

This work will be dependent on new instrumentation being built as a cowboy stream handler

PR #223, opened 2020-05-22
Improve tracing for "External" segment/spans

andrewhr:
For external traces, there is branching code to leverage the component/method/url attribute data, which maps better to how NewRelic expects this sort of workload to be reported.

This patch here expands to do the same around Segments and Spans, so transaction names follow the same patterns we see for other Agents (my baseline is Ruby). The same sort of naming pattern is found on aggregated metrics around NewRelic One UI.

A new helper function was added to `DistributedTracing` so we can, for external calls, append the HTTP status code to reported attributes. This is to also match what some agents (again, Ruby) reports on their spans for completed HTTP calls.

At last, instrumented `HTTPoison` got updated to add this HTTP status code when available.


andrewhr:
Bump to latest master

binaryseed:
This one slipped away, sorry about that! I'm going to pull this into a feature branch and make some tweaks. I'm going to include the span name changes but not the status code attribute

Thanks!

PR #225, opened 2020-05-26
Phoenix telemetry

binaryseed:
WIP.

This is intended to be merged after the plug PR, and is based on top of it's current state


lib/new_relic/telemetry/plug.ex
```diff
@@ -0,0 +1,222 @@
+defmodule NewRelic.Telemetry.Plug do
```
tpitale, line 1: Oh, curious: does this need to be a gen_server? I assume that some of things it is calling need a process as the caller or something like that. I don't think telemetry strictly requires it though, just a function, maybe? Or there's likely some other reason.

tpitale: This is similar to the other PR, I realized.

binaryseed:
Closing. Going to pull this into the main PR

PR #230, opened 2020-05-29
Fix tracer bug with struct literal and enforced_keys

### LINKED ISSUE #229: Compilation error when tracing function with NaiveDateTime
Compilation fails when attempting to trace a function that matches on NaiveDateTime.
In fact, it fails for any struct that has enforced keys

```elixir
defmodule Traced do
  use NewRelic.Tracer

  @enforce_keys [:hello]
  defstruct [:hello]

 # This doesn't compile
 @trace :enforced
 def enforced(%Traced{}), do: nil

  # Neither does this
  @trace :naive
  def naive(%NaiveDateTime{}), do: nil
end
```

**Environment**
Environment
Elixir 1.9.4 OTP22
agent in 1.18.1 (doesn't seems to be related to the fix in 1.18.2)

binaryseed:
Fixes a tracer compilation issue when an argument is a matched struct with `enforced_keys`

fixes #229

### DIFF
--- a/lib/new_relic/tracer/macro.ex
+++ b/lib/new_relic/tracer/macro.ex
@@ -223,6 +223,12 @@ defmodule NewRelic.Tracer.Macro do
     Macro.postwalk(args, &rewrite_call_term/1)
   end
 
+  # Unwrap Structs into Maps when reporting
+  # They can't always be re-referenced directly since they can have enforced_keys
+  def rewrite_call_term({:%, _, [{:__aliases__, _, _} = struct, {:%{}, line, members}]}) do
+    {:%{}, line, [{:__struct__, struct}] ++ members}
+  end
+
   # Strip default arguments
   def rewrite_call_term({:\\, _, [arg, _default]}), do: arg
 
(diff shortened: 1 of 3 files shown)


lib/new_relic/tracer/macro.ex
```diff
@@ -223,6 +223,12 @@ defmodule NewRelic.Tracer.Macro do
     Macro.postwalk(args, &rewrite_call_term/1)
   end
 
+  # Unwrap Structs into Maps when reporting
+  # They can't always be re-referenced directly since they can have enforced_keys
+  def rewrite_call_term({:%, _, [{:__aliases__, _, _} = struct, {:%{}, line, members}]}) do
```
mattbaker, line 228: Had to play around in `iex`, but to make sure I understand: this function works because the first value of that tuple would be `:%{}` if this was a map, vs `:%` when it's a struct, right?

binaryseed: Yeah, it's basically unwrapping a struct literal into plain map.

It's data that's only used for reporting so it's not changing any aspect of the function definition itself.

PR #232, opened 2020-06-04
Config docs

VitorTrin:
Improment to the configuration docs. It needed a link from the readme, somewhere that list all the options and how to use them, and the `NewRelic.Config` functions aren't used to configure, they just return the current configuration and are of little use to the consumer.


binaryseed:
Thanks for the contribution. I've reworked the config docs in this PR: #272, so I'm going to close this...

PR #235, opened 2020-06-05
Fix bug when tracing special form argument

### LINKED ISSUE #234: Compilation error when traced function has __MODULE__
I get the following error when tracing a function with `__MODULE__` in it's match

```
:__ignored__.__struct__/1 is undefined, cannot expand struct :__ignored__
```

How to reproduce

```elixir
   defmodule Traced do
     use NewRelic.Tracer

     defstruct [:a, :b]

     @trace :steve
     def steve(%__MODULE__{a: a}) do
     end
   end
```

binaryseed:
When rewriting traced functions, we need to differentiate between special forms (`__MODULE__`) and ignored variables (`_ignored`)

Fixes #234


mattbaker:
Nice catch @VitorTrin

Issue #236, opened 2020-06-08
Compilation error with enforced key


VitorTrin:
**Describe the bug**
It seems to be a corner case of the previous enforced keys problem. The problem seem to arise when you match some of the keys but not all required ones

How to reproduce:

```elixir
  defmodule Traced do
    use NewRelic.Tracer
    @enforce_keys [:key, :second_key]
    defstruct [:key, :second_key]

    @trace :mod
    def mod(%Traced{key: val}), do: val
  end
```

binaryseed:
Hi there! I took a look, but I can't reproduce this with the code provided, could you submit a PR with code changes that cause the problem you are seeing?

binaryseed:
Fix in https://github.com/newrelic/elixir_agent/pull/239

PR #238, opened 2020-06-12
Reproduce compilation error

VitorTrin:
Reproduces #236


VitorTrin:
Wait, this isn't compiling but not for bug reasons. Let me take a different approach

VitorTrin:
There it is
image

binaryseed:
ahh, the function needed to be executed with both keys, thats what i was missing. thanks!

PR #256, opened 2020-07-30
headers needs to be reassigned to include the NewRelic.distributed_trace_headers

mirjoy:
Just a small documentation update for clarity. As written the NewRelic headers are not passed to the `HttpClient`.


binaryseed:
Thanks!

PR #261, opened 2020-09-04
Add Ecto TDS adapter support

simonprev:
Just a simple implementation copied from myxql support. I did not see how can one achieve this directly in my app. Perhaps using protocol to define behaviour based on the struct should be used here but it requires some refactoring that are much bigger than my changes.

**Tests are required!**
I did not see any tests that validated the metadata parsing.

**Instrumentation particular to a library or framework are probably a better fit for their own package which depends on this Agent.**
Since the TDS adapter is officially documented in Ecto, I think that the New Relic agent should support it. Currently is raise an error on the handler and stop instrumenting.

### DIFF
--- a/lib/new_relic/telemetry/ecto/metadata.ex
+++ b/lib/new_relic/telemetry/ecto/metadata.ex
@@ -74,7 +74,7 @@ defmodule NewRelic.Telemetry.Ecto.Metadata do
         _ -> {"other", "other"}
       end
 
-    {"TDS", table, operation}
+    {"MSSQL", table, operation}
   end
 
   def parse(%{result: {:ok, %{__struct__: Postgrex.Cursor}}}), do: :ignore
--- a/lib/new_relic/telemetry/ecto/metadata.ex
+++ b/lib/new_relic/telemetry/ecto/metadata.ex
@@ -53,6 +53,30 @@ defmodule NewRelic.Telemetry.Ecto.Metadata do
     {"MySQL", table, operation}
   end
 
+  @tds_select ~r/FROM \[(?<table>\w+)\]/
+  @tds_insert ~r/INSERT INTO \[(?<table>\w+)\]/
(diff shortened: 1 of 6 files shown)


lib/new_relic/telemetry/ecto/metadata.ex
```diff
@@ -53,6 +53,30 @@ defmodule NewRelic.Telemetry.Ecto.Metadata do
...
+        "INSERT" <> _ -> {"insert", capture(@tds_insert, query, "table")}
+        "UPDATE" <> _ -> {"update", capture(@tds_update, query, "table")}
+        "DELETE" <> _ -> {"delete", capture(@tds_delete, query, "table")}
+        "begin" -> {"begin", "other"}
+        "commit" -> {"commit", "other"}
+        "rollback" -> {"rollback", "other"}
+        _ -> {"other", "other"}
+      end
+
+    {"TDS", table, operation}
```
binaryseed, line 77: The adapter is `TDS`, but the database would be `MSSQL`, right?

binaryseed (changes requested):
Thanks for this!

Tests are actually accomplished via the `examples` subdirectory, there's an `ecto_example` app that runs alongside some docker containers with the DBs running...

We'll want to get that setup since there could be other result structures to account for (take a look a the last few `parse` function heads...)

binaryseed:
Thanks for tackling this!

* Looks like a file isn't formatted
* You'll need to add to `examples/docker-compose.yml` in the examples folder so a test db is spun up. Try to keep that to a minimum if you don't mind :)

binaryseed:
you can run the tests locally pretty easily just using the same commands as the Github Action:

https://github.com/newrelic/elixir_agent/blob/master/.github/workflows/elixir.yml#L59

binaryseed:
No longer needed with: https://github.com/newrelic/elixir_agent/pull/337

PR #269, opened 2020-09-11
Avoid @trace warning when unnecessary

barthez:
Don't warn about using `@trace` on function with `rescue`, `else`, etc. if there is no `@trace` attribute set.

Running `mix test` before:
image
after:
image


AlexVPopov (approved):

binaryseed:
Thanks!

binaryseed (approved):

binaryseed:
Releasing in `v1.20.0`

PR #272, opened 2020-09-23
Logs in context

binaryseed:
This PR adds support for the [Logs In Context](https://docs.newrelic.com/docs/logs/enable-log-management-new-relic/configure-logs-context/configure-logs-context-apm-agents) feature. This feature connects `Logger` messages with the current Distributed Trace / Error Trace.

The feature is `:disabled` by default, but can be enabled in 2 modes: `direct` and `forwarder`.
* `:direct` collects logs in the agent and pushes them to New Relic at 5 second intervals
* `:forwarder` transform log messages into JSON format, which are then picked up by a [Log Forwarder](https://docs.newrelic.com/docs/logs/enable-log-management-new-relic/enable-log-monitoring-new-relic/enable-log-management-new-relic) which sends the logs to New Relic

To support this PR, there's a re-org of the harvester code and a new harvester pattern for sending data to the New Relic Telemetry APIs (to enable sending logs in `direct` mode).

Alongside this, there's a rewrite of most of the configuration docs & a performance refactor of configuration storage & fetching using `persistent_term`.


Cgboal:
Does this only capture logs emitted within a traced function?

PR #273, opened 2020-10-03
Handle missing tracestate sampled value

binaryseed:
If a `tracestate` reaches us with a missing sampled flag but it does have a priority, we need to make sure we generate a new sampling decision.

### DIFF
--- a/lib/new_relic/distributed_trace/w3c_trace_context/trace_state.ex
+++ b/lib/new_relic/distributed_trace/w3c_trace_context/trace_state.ex
@@ -131,6 +131,8 @@ defmodule NewRelic.DistributedTrace.W3CTraceContext.TraceState do
            priority,
            timestamp
          ] <- String.split(value, "-"),
+         {sampled, priority} <-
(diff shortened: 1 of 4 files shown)


lib/new_relic/distributed_trace.ex
```diff
@@ -70,20 +70,22 @@ defmodule NewRelic.DistributedTrace do
...
+    |> maybe_generate_sampled_priority()
     |> maybe_generate_trace_id()
     |> report_attributes(transport_type: type)
     |> report_attributes(:w3c)
     |> convert_to_outbound()
     |> set_tracing_context()
   end
 
-  def maybe_generate_sampling(%Context{sampled: nil, priority: nil} = context) do
+  def maybe_generate_sampled_priority(%Context{sampled: sampled, priority: priority} = context)
```
emeryotopalik, line 81: ? this fixes it?

binaryseed: The change is that this function clause will get triggered if _either_ are `nil` instead of when _both_ are nil...

PR #284, opened 2020-10-21
Transaction sidecar & Plug / Phoenix telemetry

### LINKED ISSUE #137: Not able to use recon_trace
**Describe the bug**
We occasionally use `:recon_trace.calls({Module, function, :return_trace}, n)` in production to debug issue (if the logs are not sufficient to track down the issue). With new_relic_agent enabled, this leads to following error, although the trace appears to work

> ** can only have one tracer per process

1) is there any way to fix this?
2) if not possible, is it possible to disable/enable new_relic for the duration via remote_console? I tried to stop new_relic via Application.stop(x), but it didn't go well, as all the calls to newrelic modules were failing.

**Environment**
```
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.4 (compiled with Erlang/OTP 21)
```

new_relic_agent 1.9.3

### LINKED ISSUE #259: High CPU usage when many subprocesses are spawned from a Phoenix controller
**Describe the bug**
New Relic processing takes a lot of CPU + memory when many subprocesses are spawned from a Phoenix controller.

**Environment**
  ```
  Erlang/OTP 22 [erts-10.7.2.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

  Elixir 1.10.4 (compiled with Erlang/OTP 22)
  ```
  ```
  * new_relic_agent 1.19.4 (Hex package) (mix)
    locked at 1.19.4 (new_relic_agent) 03bd88cd
  ```

**Details**
Attaching a tarball of a Phoenix app that demonstrates the problem.
[nrr.tar.gz](https://github.com/newrelic/elixir_agent/files/5152618/nrr.tar.gz)

To run, use either Docker or the raw app:
- raw app
  ```sh
  tar zxv < nrr.tar.gz
  NEW_RELIC_APP_NAME='Your app here' NEW_RELIC_LICENSE_KEY=xyz123 mix phx.server
  ```
- Docker
  ```sh
  docker build -t nr-test - < nrr.tar.gz
  docker run --rm -i -t -p 4000:4000 -e NEW_RELIC_APP_NAME='Your app here' -e NEW_RELIC_LICENSE_KEY=xyz123 nr-test
  ```

In both cases, if either `NEW_RELIC_` variable is missing, `:new_relic_agent` and `:new_relic_phoenix` are completely disabled (OTP apps don't even get loaded).

Test endpoint `/api/many/:n` (e.g., `curl localhost:4000/api/many/10000`) spawns `n` processes, each of which sleeps for 500ms (but does no other work). Params:

Name | Default | Description
:-- | :-- | :--
`n` | in path | (required) number of processes to spawn
`sleep` | 500 | number of ms to sleep in each subtask
`task_wait` | 30,000 | max number of ms to wait for each task
`overall_wait` | 30,000 | max number of ms to wait for all tasks

On my laptop, I get these results:

With New Relic disabled, these both complete successfully with no noticeable spike in CPU usage afterward:
- `/api/many/100000` returns in ~1.2s
- `/api/many/1000000` returns in ~11.8s

With New Relic enabled:
- `/api/many/100000` returns in less than 2s, but one CPU remains pegged for at least a few minutes (3+ min of ~100% CPU in htop)
- `/api/many/1000000` returns in ~23s, but two CPUs are pegged for ~1min, and another remains pegged for a while (10+ min)

### LINKED ISSUE #271: erlang tracing
While browsing through the code I noticed that the monitor agent [sets up erlang trace](https://github.com/newrelic/elixir_agent/blob/d86b04913ec58182ab0d7dbd697d816847c1aa95/lib/new_relic/transaction/monitor.ex#L25) to the instrumented process. From what I can tell, this is done for the purpose of catching spawned tasks and poolboy transactions, to make the transaction traces more detailed.

Unfortunately I feel that this approach is too aggressive. As mentioned [at the end of trace doc](http://erlang.org/doc/man/erlang.html#trace-3):

> Each process can only be traced by one tracer. Therefore, attempts to trace an already traced process fail.

This means that as a developer I can't trace instrumented process myself, which limits my options when troubleshooting the production. Moreover, the current solution feels overly naive, because there are other ways how I can start related child processes, (e.g. `Supervisor.start_child`).

Therefore, I'd like to be able to opt-out of erlang tracing. Ideally, a different kind of API would be supported which would allow clients to hook up instrumentation with some existing transaction, for example something along the lines of:

```elixir
# inside an instrumented transaction
parent = self()

Task.start_link(fn ->
  NewRelic.inside_transaction(parent)
  # ...
end)
```

This would require a bit more manual work, but on the upside it doesn't require tracing a process, and it can handle a much wider range of scenarios than currently supported.

### LINKED ISSUE #308: Very high memory usage for transaction with lots of database calls
**Describe the bug**

We're running a background job that makes 100,000s of Postgres calls inside a single process. In doing so, memory usage of the `NewRelic.Transaction.Reporter.Collecting` ETS table climbs to hundreds of megabytes, causing us to hit the memory usage limit on our Heroku dyno.

It looks like the table is filling with each individual Postgres database call, and isn't being flushed. I've attached a couple of screenshots early in the process, before we get out of memory issues.

**Environment**

* Elixir & Erlang version (`elixir -v`): `Elixir 1.11.2 (compiled with Erlang/OTP 23)`
* Agent version (`mix deps | grep new_relic_agent`): `1.23.0`

image (1)

image (2)

binaryseed:
This PR is a major refactor of the agent Transaction tracking mechanism, alongside adoption of `telemetry` based instrumentation for Plug & Phoenix!

* Introduce a `Transaction.Sidecar` process for gathering each Transaction's data which replaces the previous global ETS store
  * Cross-process Transaction tracing no longer relies on an `:erlang.trace` to connect which increases reliability and removes a bottleneck
  * Process span generation still uses `:erlang.trace` but has overload protection and can be manually disabled
* Adopts `cowboy`, `plug` and `phoenix` telemetry events for instrumentation instead of the `NewRelic.Transaction` macro
 
Additional optimizations will now be possible going forward, like putting limits on the data collected in a single Transaction.

### Upgrading

You will need to remove some old instrumentation to upgrade to this version.

All users:
* Remove `use NewRelic.Transaction`
* Remove calls to `NewRelic.Transaction.handle_errors/2`

If using `new_relic_phoenix`:
* Remove the `new_relic_phoenix` dep
* Remove `use NewRelic.Phoenix.Transaction`
* Remove `instrumenters: [NewRelic.Phoenix.Instrumenter]`

This is being published as `1.24.0-rc` for folks to try out.

Closes #137
Closes #271
Closes #259
Closes #308


bmrsny:
> This is being published as 1.23.0-rc.3 for folks to try out.

@binaryseed didn't see this tag listed above.

benizi:
> > This is being published as 1.23.0-rc.3 for folks to try out.
> 
> @binaryseed didn't see this tag listed above.

@bmrsny 
It's available via hex.pm ([full version list there](https://hex.pm/packages/new_relic_agent/versions)) (`deps` tuple needs to specify that pre-release versions are okay, e.g., `{:new_relic_agent, "~> 1.23.0-rc", runtime: false}`). But, it's still not usable where we're using it, since `:new_relic_phoenix` has a deps spec that doesn't match pre-releases.

@binaryseed 
More generally, it's odd that the set of versions on Hex doesn't match the set of tags here -- are things being published from a different repo?

binaryseed:
I'm not quite clear what you are asking about, this PR isn't merged yet - I published the pre-release from this branch..

binaryseed:
Quick note, `runtime: true` might cause problems, the agent supervision tree does indeed need to be started.

Also, `override: true` should let you work around version mismatch with other libraries.

benizi:
@binaryseed 

> I'm not quite clear what you are asking about, this PR isn't merged yet - I published the pre-release from this branch..

Other pre-release versions still seem to have been tagged in the GitHub repo, but not all.

Examples:

| version | Hex package | Git tag |
| :------ | :---------- | :------ |
| `1.23.0-rc.5` | [1.23.0-rc.5](https://hex.pm/packages/new_relic_agent/1.23.0-rc.5) | :heavy_multiplication_x:                                                           |
| `1.23.0-rc.4` | [1.23.0-rc.4](https://hex.pm/packages/new_relic_agent/1.23.0-rc.4) | :heavy_multiplication_x:                                                           |
| `1.20.0-rc.1` | [1.20.0-rc.1](https://hex.pm/packages/new_relic_agent/1.20.0-rc.1) | [v1.20.0-rc.1](https://github.com/newrelic/elixir_agent/releases/tag/v1.20.0-rc.1) |
| `1.18.0-rc.5` | [1.18.0-rc.5](https://hex.pm/packages/new_relic_agent/1.18.0-rc.5) | [v1.18.0-rc.5](https://github.com/newrelic/elixir_agent/releases/tag/v1.18.0-rc.5) |

<details>
<summary>Full list</summary>

| version | Hex package | Git tag |
| :------ | :---------- | :------ |
| `1.23.0-rc.5` | [1.23.0-rc.5](https://hex.pm/packages/new_relic_agent/1.23.0-rc.5) | :heavy_multiplication_x:                                                           |
| `1.23.0-rc.4` | [1.23.0-rc.4](https://hex.pm/packages/new_relic_agent/1.23.0-rc.4) | :heavy_multiplication_x:                                                           |
| `1.23.0-rc.3` | [1.23.0-rc.3](https://hex.pm/packages/new_relic_agent/1.23.0-rc.3) | :heavy_multiplication_x:                                                           |
| `1.23.0-rc.2` | [1.23.0-rc.2](https://hex.pm/packages/new_relic_agent/1.23.0-rc.2) | :heavy_multiplication_x:                                                           |
| `1.23.0-rc.1` | [1.23.0-rc.1](https://hex.pm/packages/new_relic_agent/1.23.0-rc.1) | :heavy_multiplication_x:                                                           |
| `1.21.0-rc.3` | [1.21.0-rc.3](https://hex.pm/packages/new_relic_agent/1.21.0-rc.3) | :heavy_multiplication_x:                                                           |
| `1.21.0-rc.1` | [1.21.0-rc.1](https://hex.pm/packages/new_relic_agent/1.21.0-rc.1) | :heavy_multiplication_x:                                                           |
| `1.20.0-rc.1` | [1.20.0-rc.1](https://hex.pm/packages/new_relic_agent/1.20.0-rc.1) | [v1.20.0-rc.1](https://github.com/newrelic/elixir_agent/releases/tag/v1.20.0-rc.1) |
| `1.18.0-rc.5` | [1.18.0-rc.5](https://hex.pm/packages/new_relic_agent/1.18.0-rc.5) | [v1.18.0-rc.5](https://github.com/newrelic/elixir_agent/releases/tag/v1.18.0-rc.5) |
| `1.18.0-rc.4` | [1.18.0-rc.4](https://hex.pm/packages/new_relic_agent/1.18.0-rc.4) | [v1.18.0-rc.4](https://github.com/newrelic/elixir_agent/releases/tag/v1.18.0-rc.4) |
| `1.18.0-rc.3` | :heavy_multiplication_x:                                           | [v1.18.0-rc.3](https://github.com/newrelic/elixir_agent/releases/tag/v1.18.0-rc.3) |
| `1.18.0-rc.2` | [1.18.0-rc.2](https://hex.pm/packages/new_relic_agent/1.18.0-rc.2) | [v1.18.0-rc.2](https://github.com/newrelic/elixir_agent/releases/tag/v1.18.0-rc.2) |
| `1.18.0-rc.1` | [1.18.0-rc.1](https://hex.pm/packages/new_relic_agent/1.18.0-rc.1) | [v1.18.0-rc.1](https://github.com/newrelic/elixir_agent/releases/tag/v1.18.0-rc.1) |
| `1.17.0-rc.4` | [1.17.0-rc.4](https://hex.pm/packages/new_relic_agent/1.17.0-rc.4) | :heavy_multiplication_x:                                                           |
| `1.17.0-rc.3` | [1.17.0-rc.3](https://hex.pm/packages/new_relic_agent/1.17.0-rc.3) | [v1.17.0-rc.3](https://github.com/newrelic/elixir_agent/releases/tag/v1.17.0-rc.3) |
| `1.17.0-rc.2` | [1.17.0-rc.2](https://hex.pm/packages/new_relic_agent/1.17.0-rc.2) | [v1.17.0-rc.2](https://github.com/newrelic/elixir_agent/releases/tag/v1.17.0-rc.2) |
| `1.17.0-rc.1` | [1.17.0-rc.1](https://hex.pm/packages/new_relic_agent/1.17.0-rc.1) | [v1.17.0-rc.1](https://github.com/newrelic/elixir_agent/releases/tag/v1.17.0-rc.1) |
| `1.17.0-rc.0` | [1.17.0-rc.0](https://hex.pm/packages/new_relic_agent/1.17.0-rc.0) | [v1.17.0-rc.0](https://github.com/newrelic/elixir_agent/releases/tag/v1.17.0-rc.0) |
| `1.16.0-rc.2` | [1.16.0-rc.2](https://hex.pm/packages/new_relic_agent/1.16.0-rc.2) | :heavy_multiplication_x:                                                           |
| `1.16.0-rc.1` | [1.16.0-rc.1](https://hex.pm/packages/new_relic_agent/1.16.0-rc.1) | :heavy_multiplication_x:                                                           |
| `1.16.0-rc.0` | [1.16.0-rc.0](https://hex.pm/packages/new_relic_agent/1.16.0-rc.0) | :heavy_multiplication_x:                                                           |
| `1.15.0-rc.3` | [1.15.0-rc.3](https://hex.pm/packages/new_relic_agent/1.15.0-rc.3) | :heavy_multiplication_x:                                                           |
| `1.15.0-rc.2` | [1.15.0-rc.2](https://hex.pm/packages/new_relic_agent/1.15.0-rc.2) | [v1.15.0-rc.2](https://github.com/newrelic/elixir_agent/releases/tag/v1.15.0-rc.2) |
| `1.15.0-rc.1` | [1.15.0-rc.1](https://hex.pm/packages/new_relic_agent/1.15.0-rc.1) | [v1.15.0-rc.1](https://github.com/newrelic/elixir_agent/releases/tag/v1.15.0-rc.1) |

</details>

> Quick note, `runtime: true` might cause problems, the agent supervision tree does indeed need to be started.

We use `runtime: false` and start it manually in our own supervision tree depending on Application environment. (Added the ability to disable it without entirely ripping it out when we hit #259)

> Also, `override: true` should let you work around version mismatch with other libraries.

`override: true` didn't seem to be sufficient for `:new_relic_phoenix` to accept the pre-release version. (Though, I'll prob. attempt that again at this point.)

Nitrino:
@binaryseed Do you have any progress with this PR?

forest:
Thanks for pushing this forward. I did some local testing with a phoenix app and found that `@plug_router_start [:plug, :router_dispatch, :start]` is never called resulting in the transaction name being `WebTransaction/Unknown/Unknown`.

If I update `@plug_router_start` to `[:phoenix, :router_dispatch, :start]` then I get a plug_name. `plug_name: "/Plug/GET//controller/resource/:id"`.

Maybe adding another handle_event for `[:phoenix, :router_dispatch, :start]` is an initial solution.

binaryseed:
FYI I had been working on getting Phoenix instrumentation wrapped into this, and I just pushed that up

binaryseed:
FYI, released an Agent version with another feature, so pushing this to `1.24.0-rc` instead of `1.23.0` shortly..

lib/new_relic/telemetry/phoenix.ex
```diff
@@ -0,0 +1,94 @@
...
+      "phoenix.router": conn.private[:phoenix_router] |> inspect()
+    ]
+    |> NewRelic.add_attributes()
+  end
+
+  def handle_event(_event, _measurements, _meta, _config) do
+    :ignore
+  end
+
+  defp phoenix_name(%{plug: controller, plug_opts: action}) do
```
derrick-zml, line 83:
```suggestion
  defp phoenix_name(%{plug: controller, plug_opts: action}) when is_atom(action) do
```

The `plug_opts` may be a value cannot convert to string. For example, if we have `Absinthe.Plug` in the router like this:

```elixir
  forward "/graphql", Absinthe.Plug, schema: MyAppWeb.Schema
```

`plug_opts` will be `[schema: MyAppWeb.Schema`

lib/new_relic/telemetry/phoenix.ex
```diff
@@ -0,0 +1,94 @@
...
+        _measurements,
+        %{conn: conn} = meta,
+        _config
+      ) do
+    [
+      phoenix_name: phoenix_name(meta),
+      "phoenix.endpoint": conn.private[:phoenix_endpoint] |> inspect(),
+      "phoenix.router": conn.private[:phoenix_router] |> inspect(),
+      "phoenix.controller": meta.plug |> inspect(),
+      "phoenix.action": meta.plug_opts |> to_string(),
```
derrick-zml, line 57: This will raise error when `plug_opts` is a keywork list.

lib/new_relic/telemetry/phoenix.ex
zillou, line 83:
```suggestion
  defp phoenix_name(%{plug: controller, plug_opts: action}) when is_atom(action) do
```

The `plug_opts` may be a value cannot convert to string. For example, if we have `Absinthe.Plug` in the router like this:

```elixir
  forward "/graphql", Absinthe.Plug, schema: MyAppWeb.Schema
```

`plug_opts` will be `[schema: MyAppWeb.Schema`

CHANGELOG.md
```diff
@@ -15,6 +28,7 @@ Tweaks
 
 Tweaks
 * Improve External span names. [#309](https://github.com/newrelic/elixir_agent/pull/309)
+>>>>>>> master
```
zillou, line 31:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+  * Please remove deprecated calls:
+    - `use NewRelic.Transaction`
+    - `NewRelic.Transaction.handle_errors/2`
+* `Phoenix` instrumentation is not fullly automatic based on `telemetry` events!
+  * Please remove deprecated instrumentation library:
+    - https://github.com/binaryseed/new_relic_phoenix
+* Transaction tracking is now faster and better in the face of overload
+  * Transactions no longer follow processes that aren't linked (ex: `Task.Supervisor.async_nolink`). They can be connected manually if desired using `NewRelic.connect_task_to_transaction` and `NewRelic.connect_to_transaction`.
+
+=======
```
joaothallis, line 16:
```suggestion
```

binaryseed: Thanks!

PR #285, opened 2020-10-26
Add runtime_tools to extra_applications

tomtaylor:
This prevents a compilation warning on Elixir 1.11 about using a dependency that hasn't been defined.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -16,6 +16,10 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


binaryseed:
Ok, thanks for the PR but rather than always start up `runtime_tools`, I've just implemented the util calculation, it's just a little bit of math :)

#288

binaryseed:
Fixed in `1.22.1`

tomtaylor:
Thanks!

PR #289, opened 2020-10-28
Only get CPU utilization if os_mon's cpu_sup is enabled

tomtaylor:
This ensures that we return nil, not 0, if cpu_sup is disabled, and prevents endless log warnings about cpu_sup not being enabled.

### DIFF
--- a/lib/new_relic/sampler/beam.ex
+++ b/lib/new_relic/sampler/beam.ex
@@ -15,7 +15,8 @@ defmodule NewRelic.Sampler.Beam do
     :erlang.system_flag(:scheduler_wall_time, true)
 
     # throw away first value
-    :cpu_sup.util()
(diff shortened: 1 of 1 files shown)


binaryseed:
Are you manually turning off `cpu_sup`?. The agent includes the `:os_mon` application...

tomtaylor:
Yeah, that's right.

binaryseed:
Is it just the shutdown bug you are trying to avoid or is there another reason you don't want it started?

tomtaylor:
No, it's just the shutdown bug. If we disable it in the `os_mon` config ourselves the agent will still make calls to `cpu_util`, littering our logs with warnings about it not being started every few seconds.

tomtaylor:
Hey @binaryseed, any more thoughts on this?

PR #315, opened 2021-02-23
Inspect failed output going to log

binaryseed:
When bad data has wound up in the payload, we need to safely log it


binaryseed:
Followup to https://github.com/newrelic/elixir_agent/pull/295

PR #323, opened 2021-03-09
Update examples

binaryseed:
This PR includes a `test_support` app in the examples umbrella so that we don't get recompilation warnings in each test run


binaryseed:
This is what we saw before this PR:
Screen Shot 2021-03-09 at 11 45 44 AM

PR #324, opened 2021-03-10
Fix logs in context region URL

mattgibson:
The region extracted from licence keys such as `eu01xx________` is `eu01`, which works for the tracing, but not for logging in context. The [logging API docs](https://docs.newrelic.com/docs/logs/log-management/log-api/introduction-log-api/#endpoint) require `https://log-api.eu.newrelic.com/log/v1`, but currently, the URL ends up as `https://log-api.eu01.newrelic.com/log/v1` which fails to report any data.


binaryseed:
Thanks, I think both of the URLs actually need to change so I'm going to tweak this and then merge into master

PR #325, opened 2021-03-10
Add the comments from the logging in context configration code to the README

mattgibson:
I didn't realise the agent supported this feature and had to go on safari through the code to work out how it could be configured.

This PR adds the existing configuration instructions to the README so it's clear that this can be done.


binaryseed:
Hi, thanks for this. I'm going to tack on a commit to #326 with a note about finding this easier rather than adding all of this in two places...

PR #326, opened 2021-03-10
Fix Telemetry API URLS for EU

binaryseed:
This PR fixes the Telemetry API URLs for the EU region

Continued from #324


binaryseed:
Releasing in `v1.24.3`

PR #328, opened 2021-03-14
Include metadata which has been added with a specific log message

mattgibson:
Previously only whole-process metadata added via `Logger.metadata()` would be picked up. Now calls like `Logger.error("FOO", baz: :qux)` will send `"metadata.bax": "qux"` to New Relic, whereas previously that would be missing.


binaryseed:
Thanks, merging into a feature branch to make some small tweaks

binaryseed:
Released in `1.24.4`

PR #329, opened 2021-03-14
Handle IO lists so that the new relic logs show strings, rather than lists in the UI

mattgibson:
Before:

Screenshot 2021-03-14 at 23 18 20

After:

Screenshot 2021-03-14 at 23 18 30


binaryseed:
Thanks!

binaryseed:
Released in `1.24.4`

PR #330, opened 2021-03-15
Support per-log message metadata

binaryseed:
This PR adds support for per-log message metadata.

Continued from #328


binaryseed:
Released in `1.24.4`

PR #336, opened 2021-04-05
Support sqlite3 and oracle adapter

marocchino:
Hi, Our company recently started using New Relic.
After setting and checking the log, I found that this library does not support our DB and a matching error occurs.

This PR skips when a matching error occurs and adds support for sqlite3 and Oracle.

I hope they are merged as soon as possible.
Please tell me if you have any problems.


lib/new_relic/telemetry/ecto/metadata.ex
```diff
@@ -1,16 +1,48 @@
...
+
+  def parse(%{
+        query: query,
+        result: {_ok_or_error, %{__struct__: struct}}
+      })
+      when struct in [Exqlite.Result, Exqlite.Error] do
+    parse_query(query, "SQLite3")
+  end
+
+  def parse(%{query: query, repo: repo, result: {_ok_or_error, _}}) do
```
marocchino, line 28: This part has not been tested due to several technical difficulties:
- Docker images of Oracle instances require license agreement.
- jamdb_oracle does [not support](https://github.com/erlangbureau/jamdb_oracle/blob/d5c87dd04fdf2206d27f09a13144a115764ee429/lib/jamdb_oracle_ecto.ex#L189-L193) storage_down storage_up.
- The volume size of an Oracle instance is about 3 gigabytes. If we test every time, it takes a lot of time to initialize.

lib/new_relic/telemetry/ecto/metadata.ex
```diff
@@ -1,16 +1,49 @@
...
+    case repo.__adapter__() do
+      Ecto.Adapters.Jamdb.Oracle -> parse_query(query, "Oracle")
+      _ -> :ignore
+    end
+  end
+
+  def parse(%{result: {:ok, %{__struct__: Postgrex.Cursor}}}), do: :ignore
+  def parse(%{result: {:ok, %{__struct__: MyXQL.Cursor}}}), do: :ignore
+  def parse(%{result: {:ok, %{__struct__: SQLite3.Cursor}}}), do: :ignore
+  def parse(%{result: {:ok, nil}}), do: :ignore
```
marocchino, line 38: Although not included in this PR, a matching error may still occur in this part. Can I modify it like this? 
```suggestion
  def parse(%{result: {:ok, _}}), do: :ignore
```

binaryseed:
Thanks for this!

I took a look on Hex and I'm finding that usage of these two adapters is very low, almost 1000x lower than the common postgres...

Given that, I'm thinking that it might be best to actually implement this in a more generic way instead.. ie: have a `parse` function for any "other" adapter besides the main ones we already support, and grab the name for metrics directly from the result struct.

For testing this, the existing sqlite setup you put together will still work great

I'm going to merge this into a working branch and make those changes...

binaryseed:
Will be included in https://github.com/newrelic/elixir_agent/pull/337

PR #340, opened 2021-04-06
Fix parsing of other SELECTs

binaryseed:
Fix an edge case parsing the table name out of queries like `SELECT my_lock()`. Also includes a bit more code cleanup in this area.


tomtaylor:
Thanks!

PR #342, opened 2021-04-08
Handle exception when inspecting remote process

benhaney:
When a process is spawned on a remote node, the transaction monitor tries to get its name by calling `Process.info`, which always raises badarg when the Pid given to it is non-local. This results in the transaction monitor terminating, when it's probably more desirable to just continue without a name.


binaryseed (approved):
Thanks!

binaryseed:
Released in `1.25.3`

PR #345, opened 2021-05-04
Add HTTPoison as optional dependency

andrewhr:
We have a few libraries wrapping HTTPoison that optionally instrument if NewRelic is present during compilation. Due the lack of this optional dependency here, the compilation process is not deterministic and may end up with cases where both libraries are included (our wrapper and NewRelic) and still have no instrumentation.


binaryseed:
Mind sharing a snipped of code that demonstrates how you are optionally instrumenting?

binaryseed:
Would you mind sharing an example of how you are instrumenting so I can think about this one?

chazsconi:
Is it possible re-open this and apply the fix in the PR to make `httpoison` and optional dependency? 

I don't believe it is possible to use the instrumented wrapper around this library (https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/instrumented/httpoison.ex) without this fix as the wrapper will never get compiled due the code `if Code.ensure_loaded?(HTTPoison) do` at line 1.

Please see this repo for a minimal example of the compile time warning (that will become a runtime error) that currently exists: https://github.com/chazsconi/new_relic_agent_test

Running `mix deps.get` followed by `mix compile` will show this:

```
Generated new_relic_agent app
    warning: NewRelic.Instrumented.HTTPoison.request/5 is undefined (module NewRelic.Instrumented.HTTPoison is not available or is yet to be defined)
    │
  5 │     do: HTTPoison.request(method, url, body, headers, options)
    │                   ~
    │
    └─ lib/new_relic_test.ex:5:19: NewRelicTest.request/5
```

Issue #260, opened 2020-09-01
Support for high security mode 


codebender:
Hi!
I was looking through the code base for "high security mode" but I don't think I found anything.

Is there a reason this agent doesn't support high security mode like your other agents?

binaryseed:
This agent is released as open-source, which means it doesn't support everything New Relic as a whole supports.

That said, you can turn off security sensitive data collection including error tracing and query collection:

https://hexdocs.pm/new_relic_agent/NewRelic.Config.html#feature?/1-security

PR #346, opened 2021-05-06
Add support for HSM v2

tyler-gibbons:
The Elixir New Relic Agent doesn't currently support High Security Mode (HSM) in any shape or form. Ideally the setting is not enabled on the NR account and we get the full scope and breadth of reported metrics and spans, but..

Our New Relic account at Afterpay is configured to use HSM and, given that our services are built to assume this, the proposition of disabling HSM account-wide is incredibly dicey. The New Relic agent would immediate get forcefully disconnected because the NR server would assume we were an insecure agent under (deprecated) [HSMv1 requirements](https://docs.newrelic.com/docs/agents/manage-apm-agents/configuration/high-security-mode/#version1description). This was a blocking issue for setting up metrics reporting for the Elixir service I maintain here, but we found that [we could upgrade metrics reporting to work with HSMv2](https://docs.newrelic.com/docs/agents/manage-apm-agents/configuration/high-security-mode/#migrating) by monkeypatching the `NewRelic.Harvest.Collector.Connect` module to include a `high_security: true` parameter in the payload.

Once under HSMv2, the agent is allowed to continue to connect and any potentially sensitive information is stripped out on the New Relic side before showing up in collected metrics.

This PR formalizes that into an actual configuration setting and should have no impact for any New Relic agents reporting to non-HSM accounts. I would like to do further work to support high security mode-- the current implementation seems to strip a lot of useful information present in, e.g., Python-reported metrics to the same account-- but the actual protocol/reporting-level implications of HSM aren't documented anywhere that I can find. 

At the very least, adding this configuration should enable anyone blocked on #260 to use the NR agent with an Elixir service. Please let me know if there is anything further y'all would like me to do before merging this; I'd love to be able to remove the kind-of-hacky module replacement I'm doing within my own service and any other Elixir services other teams here might set up :)

### DIFF
--- a/lib/new_relic/config.ex
+++ b/lib/new_relic/config.ex
@@ -97,6 +97,28 @@ defmodule NewRelic.Config do
     do: get(:labels)
 
   @doc """
+  Enables High Security Mode; when enabled this prevents any potentially
+  sensitive information being reported to New Relic. You can enable this
+  behavior in two ways:
+
+  * Environment variables `NEW_RELIC_HIGH_SECURITY=true`
+  * Application config: `config :new_relic_agent, high_security: true`
+
+  The configuration setting here must match your New Relic account setting;
+  otherwise the agent's connection attempt will be rejected and the agent will
+  shut down. This will prevent the agent from reporting any metrics to New Relic,
+  but will not shut down your application.
+
+  Enabling High Security Mode has significant implications for all applications
+  reporting to your New Relic account. Please read the High Security Mode
+  documentation for further details:
+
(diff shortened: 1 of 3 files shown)


binaryseed:
Unfortunately, HSM must be completely implemented in the agent itself, and there's quite a few places where restrictions must be applied to meet the criteria. Sending in this attribute is not sufficient to actually filter out sensitive data.

If you are OK with that, bypassing it yourself is your choice, but it's not safe to expose this configuration & not do the data filtering in the agent.

tyler-gibbons:
Makes sense! Is there documentation somewhere that discusses where all these restrictions need to be put in?

PR #359, opened 2021-07-27
Ensure Task is loaded before running function_exported?/3

mhanberg:
There can be a race condition when the compiler is evaluating these
lines of code, that allow for function_exported? to return false. This
is because the Task module hasn't been loaded yet.

Running `Code.ensure_loaded?(Task)` before calling `function_exported?`
will make sure that we avoid the race condition.


binaryseed:
Thanks!

PR #365, opened 2021-09-29
Update Phoenix dependency to allow Phoenix 1.6.

jimsynz:
Hey folks.

This PR loosens the Phoenix dependency to allow Phoenix to be upgraded to 1.6 while `new_relic_agent` is installed.

Thanks!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -48,7 +48,7 @@ defmodule NewRelic.Mixfile do
       # Instrumentation:
       {:plug, ">= 1.10.4", optional: true},
       {:plug_cowboy, ">= 2.4.0", optional: true},
-      {:phoenix, ">= 1.5.5", optional: true},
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -48,7 +48,7 @@ defmodule NewRelic.Mixfile do
       # Instrumentation:
       {:plug, ">= 1.10.4", optional: true},
       {:plug_cowboy, ">= 2.4.0", optional: true},
-      {:phoenix, ">= 1.5.5", optional: true},
+      {:phoenix, ">= 1.5.5 or ~> 1.6.0", optional: true},
```
binaryseed, line 51: It's not clear to me why this would be needed, the existing version requirement should match:

```elixir
iex> Version.match?("1.6.0", ">= 1.5.5")
true
```

Are you encountering an error?

jimsynz: You're right. My bad! It was caused by another dependency.  I'll close this PR.

PR #373, opened 2021-12-09
use `System.pid/0`

fishtreesugar:
instead of `System.get_pid/0`

### DIFF
--- a/lib/new_relic/util.ex
+++ b/lib/new_relic/util.ex
@@ -7,7 +7,7 @@ defmodule NewRelic.Util do
     maybe_heroku_dyno_hostname() || get_hostname()
   end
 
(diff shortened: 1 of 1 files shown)


binaryseed (approved):

MohamedBrary:
Just pointing that `System.pid/0` was introduced in Elixir 1.9.0: https://hexdocs.pm/elixir/1.12/System.html#pid/0
While the documentation for requirements for installation still points to `Elixir 1.8`
https://github.com/newrelic/elixir_agent#installation

Any earlier Elixir version would get this error:
```
GenServer NewRelic.Harvest.Collector.AgentRun terminating\n** (UndefinedFunctionError) function System.pid/0 is undefined or private
```

PR #376, opened 2022-01-14
WIP: Dimensional metric reporting

tpitale:
* Adds `NewRelic.report_dimensional_metric/4`

### DIFF
--- a/lib/new_relic.ex
+++ b/lib/new_relic.ex
@@ -230,6 +230,17 @@ defmodule NewRelic do
     to: NewRelic.Harvest.Collector.CustomEvent.Harvester
 
   @doc """
+  Report a Dimensional Metric.
+  Valid types: `:count`, `:gauge`, and `:summary`.
+
+  ```elixir
+  NewRelic.report_dimensional_metric(:count, "my_metric_name", 1, %{some: "attributes"})
+  ```
+  """
+  defdelegate report_dimensional_metric(type, name, value, attributes \\ %{}),
+    to: NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester
+
+  @doc """
   Report a Custom metric.
 
   ```elixir
--- a/lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
+++ b/lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
@@ -1,0 +1,87 @@
+defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
+  use GenServer
+
+  @moduledoc false
+
+  alias NewRelic.Harvest
+  alias NewRelic.Harvest.TelemetrySdk
+
+  @interval_ms 5_000
+
+  @valid_types [:count, :gauge, :summary]
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [])
+  end
+
+  def init(_) do
+    {:ok,
+     %{
+       start_time_ms: System.system_time(:millisecond),
+       metrics: []
+     }}
+  end
+
+  # API
+
+  @spec report_dimensional_metric(:count | :gauge | :summary, atom() | binary(), any, map()) ::
+          :ok
+  def report_dimensional_metric(type, name, value, attributes) when type in @valid_types do
+    TelemetrySdk.DimensionalMetrics.HarvestCycle
+    |> Harvest.HarvestCycle.current_harvester()
+    |> GenServer.cast({:report, %{type: type, name: name, value: value, attributes: attributes}})
+  end
+
+  def gather_harvest,
+    do:
+      TelemetrySdk.DimensionalMetrics.HarvestCycle
+      |> Harvest.HarvestCycle.current_harvester()
+      |> GenServer.call(:gather_harvest)
+
+  # do not accept more report messages when harvest has already been reported
+  def handle_cast(_late_msg, :completed), do: {:noreply, :completed}
+
+  def handle_cast({:report, metric}, state) do
+    # TODO: merge metrics with the same type/name/attributes?
+    {:noreply, %{state | metrics: [metric | state.metrics]}}
+  end
+
+  # do not resend metrics when harvest has already been reported
+  def handle_call(_late_msg, _from, :completed), do: {:reply, :completed, :completed}
+
+  def handle_call(:send_harvest, _from, state) do
+    send_harvest(state)
+    {:reply, :ok, :completed}
+  end
+
+  def handle_call(:gather_harvest, _from, state) do
+    {:reply, build_dimensional_metric_data(state.metrics, state), state}
+  end
+
+  # Helpers
+
+  defp send_harvest(state) do
+    TelemetrySdk.API.log(build_dimensional_metric_data(state.metrics, state))
+    log_harvest(length(state.metrics))
+  end
+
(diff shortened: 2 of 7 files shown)


lib/new_relic.ex
```diff
@@ -229,6 +229,17 @@ defmodule NewRelic do
...
 
+  @doc """
+  Report a Dimensional Metric.
+  Valid types: `:count`, `:gauge`, and `:summary`.
+
+  ```elixir
+  NewRelic.report_dimensional_metric(:count, "my_metric_name", 1, %{some: "attributes"})
+  ```
+  """
+  defdelegate report_dimensional_metric(type, name, value, attributes \\ %{}),
```
tpitale, line 240: I feel like there should be other functions that call this, as well. Similar to how incrementing a timeslice metric works.

lib/new_relic.ex
```diff
@@ -229,6 +229,17 @@ defmodule NewRelic do
   defdelegate report_custom_event(type, attributes),
     to: NewRelic.Harvest.Collector.CustomEvent.Harvester
 
+  @doc """
+  Report a Dimensional Metric.
+  Valid types: `:count`, `:gauge`, and `:summary`.
```
tpitale, line 234: Probably should add some docs (or a link) on when to use each of the types.

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,87 @@
+defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
+  use GenServer
+
+  @moduledoc false
+
+  alias NewRelic.Harvest
+  alias NewRelic.Harvest.TelemetrySdk
+
+  @interval_ms 5_000
```
tpitale, line 9: I set this to match the harvest cycle time. It's required for `count` and `summary` types in the `common` field.

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,87 @@
...
+    {:ok,
+     %{
+       start_time_ms: System.system_time(:millisecond),
+       metrics: []
+     }}
+  end
+
+  # API
+
+  @spec report_dimensional_metric(:count | :gauge | :summary, atom() | binary(), any, map()) ::
```
tpitale, line 27: Should I remove the typespec?

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,87 @@
...
+  defp build_dimensional_metric_data(metrics, state) do
+    [
+      %{
+        metrics: metrics,
+        common: common(state)
+      }
+    ]
+  end
+
+  defp common(%{start_time_ms: start_time_ms}) do
```
tpitale, line 84: I made the start time the common timestamp for any metrics in this harvest cycle. I don't think there is any accuracy below (or even near to) 5 seconds when querying.

test/dimensional_metric_test.exs
```diff
@@ -0,0 +1,28 @@
...
+    NewRelic.report_dimensional_metric(:summary, "memory.foo_bar", 50, %{cpu: 2000})
+
+    [%{common: common, metrics: metrics}] =
+      TestHelper.gather_harvest(NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester)
+
+    assert common["interval.ms"] > 0
+    assert common["timestamp"] > 0
+
+    assert length(metrics) == 2
+    [metric1, metric2] = metrics
```
tpitale, line 19: This points out that the metrics are coming in reverse order because that's how they're added. Not sure if it is worth swapping the order, or reversing the metrics at gather time.

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,87 @@
...
+       start_time_ms: System.system_time(:millisecond),
+       metrics: []
+     }}
+  end
+
+  # API
+
+  @spec report_dimensional_metric(:count | :gauge | :summary, atom() | binary(), any, map()) ::
+          :ok
+  def report_dimensional_metric(type, name, value, attributes) when type in @valid_types do
```
binaryseed, line 29: The main risk I'm noticing in this PR is that `report_dimensional_metric` collects all the metrics and sends every one up, while `report_custom_metric ` will aggregate each metric during the harvest cycle and only sends each one once.

Which behavior do you want this function to have? I suspect most folks would assume the aggregated behavior. For example, if someone calls this function for every single request, we wouldn't want to copy the name & attributes potentially thousands of times a second..

If you did want that behavior, perhaps `send_dimensional_metric` is more expressive of what will happen?

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,87 @@
...
+    [
+      %{
+        metrics: metrics,
+        common: common(state)
+      }
+    ]
+  end
+
+  defp common(%{start_time_ms: start_time_ms}) do
+    %{"timestamp" => start_time_ms, "interval.ms" => @interval_ms}
```
binaryseed, line 85: Probably worth pulling `"interval.ms"` from the configuration vs the module attribute in case it gets configured differently

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
binaryseed: I feel like a pretty detailed spec would be worth it on the function inside `NewRelic` module, I'd go beyond what's here to help folks understand what the `value` should be for the various types

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
tpitale: Yeah. I definitely agree that it's not what I would expect.

What are your thoughts on efficiently storing the map of attributes along with the metric name?

Eg I report metric_a with attributes a, b, and c. Then I report metric_a with attributes c, d, and e.

Do I merge the maps? Or does that remain as two metrics?

tpitale: I could use something like phash2 on the attributes map and then aggregate matching ones.

binaryseed: I think that's right, the whole of the attributes makes up a unique metric. The other metric harvester has an optimized aggregation implementation you could steal

tpitale: So it would be keyed off of a tuple of the name and the hash of the attributes? The timeslice metric harvester uses the name and the scope, it seems.

binaryseed: Yep!

tpitale:
Closing in favor of: https://github.com/newrelic/elixir_agent/pull/408

PR #378, opened 2022-02-07
Changes default custom event per minute to 10_000

dmccown:
This changes the default custom event per minutes from 1000 to 10000 to
match the [Java agent](https://docs.newrelic.com/docs/data-apis/custom-data/custom-events/apm-report-custom-events-attributes/#java-att) defaults.

### DIFF
--- a/lib/new_relic/config.ex
+++ b/lib/new_relic/config.ex
@@ -218,7 +218,8 @@ defmodule NewRelic.Config do
       harvest_limits: %{
         analytic_event_data:
           Application.get_env(:new_relic_agent, :analytic_event_per_minute, 1000),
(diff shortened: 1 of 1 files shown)


binaryseed:
Thanks, but we'll keep the default as-is, you can configure this manually if you like.

dmccown:
I'm curious on reasoning for defaulting to 1000?

 It was quite confusing on an application I was looking at where events were being dropped. We had assumed based on the documentation that we were at 10,000 as in the public New Relic docs. It took some time debugging this to figure out we were dropping events as our default was 1000.

tpitale:
I'm also curious of the history of this (David is a coworker of mine at NR @binaryseed, joined after you).

PR #389, opened 2022-08-15
Fix bug in top_process.ex when complex args are in use

Cgboal:
Hi, 

I ran into a bug when spawning supervised tasks via `Task.async_stream_nolink/6`. When spawning tasks with MFA, the TopProcess GenServer will crash when trying to parse the spawned Tasks info if the args passed to it cannot be simply converted to a string during interpolation. 

In my case, this occured when passing `[[String.t()], map()]`, which resulted in the following error: 
```
(ArgumentError) cannot convert the given list to a string.

To be converted to a string, a list must either be empty or only
contain the following elements:

  * strings
  * integers representing Unicode code points
  * a list containing one of these three elements

Please check the given list or call inspect/1 to get the list representation, got:
```

This fix simply adds an inspect call within the string interpolation which prevents this issue. I have tested this and it resolves the issue.

### DIFF
--- a/lib/new_relic/sampler/top_process.ex
+++ b/lib/new_relic/sampler/top_process.ex
@@ -76,7 +76,7 @@ defmodule NewRelic.Sampler.TopProcess do
   defp parse(:name, pid, []) do
     with {:dictionary, dictionary} <- Process.info(pid, :dictionary),
          {m, f, a} <- Keyword.get(dictionary, :"$initial_call") do
-      "#{inspect(m)}.#{f}/#{a}"
+      "#{inspect(m)}.#{f}/#{inspect(a)}"
     else
       _ -> inspect(pid)
     end
--- a/lib/new_relic/util/http.ex
+++ b/lib/new_relic/util/http.ex
@@ -42,6 +42,8 @@ defmodule NewRelic.Util.HTTP do
   https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/ssl
   """
   def http_options(opts \\ []) do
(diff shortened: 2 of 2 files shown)


lib/new_relic/sampler/top_process.ex
```diff
@@ -76,7 +76,7 @@ defmodule NewRelic.Sampler.TopProcess do
   defp parse(:name, pid, []) do
     with {:dictionary, dictionary} <- Process.info(pid, :dictionary),
          {m, f, a} <- Keyword.get(dictionary, :"$initial_call") do
-      "#{inspect(m)}.#{f}/#{a}"
+      "#{inspect(m)}.#{f}/#{inspect(a)}"
```
binaryseed, line 79: Something else must be going wrong here. The `a` in `$initial_call` is set by `Task` and is the _arity_ - it is always a number.

https://github.com/elixir-lang/elixir/blob/fb729784e5504f499b0179c4f154e73ea2a6f216/lib/elixir/lib/task/supervised.ex#L77-L85

Are you doing something manually with `$initial_call`?

Cgboal: Nope not doing anything manually with initial call:
```
    task_stream =
      Task.Supervisor.async_stream_nolink(
        XXXX.TaskSupervisor,
        batches,
        callback_module,
        :execute,
        [refined_context],
        max_concurrency: callback_module.max_concurrency(),
        timeout: :infinity
      )
```

binaryseed:
It's been a while. I'm going to close this since I don't see how `$initial_call` could ever have the args in it. Feel free to share more context from error logs if you (or anyone) have seen this again

PR #393, opened 2022-09-17
Update README.md

ahsandar:
Add docs to define httpc vars via config


mattbaker:
Merged a slightly different version this morning, so this is addressed in #394

PR #398, opened 2023-02-16
Show both LiveView and action name in transaction when action is given

goto-engineering:
We were having an issue where transactions were very confusingly labeled on LiveViews. Consider the example.

```
live "/stuff", StuffLive, :home
live "/more", MoreLive, :home
live "/home", HomeLive, :do_things
```

Because of the naming logic in the Phoenix tracing, these would log as the following:
```
StuffLive -> Phoenix.LiveView.Plug/home
MoreLive -> Phoenix.LiveView.Plug/home
HomeLive -> Phoenix.LiveView.Plug/do_things
```

If you leave the action off completely, it would just log the name of the LiveView but with `Elixir.` prefixed (fully qualified atom name) which was much nicer.

This change uses both the name of the LiveView as well as the action name if both are given, and also strips the `Elixir.`:
```
StuffLive -> Phoenix.LiveView.Plug/ProjectName.StuffLive/home
MoreLive -> Phoenix.LiveView.Plug/ProjectName.MoreLive/home
HomeLive -> Phoenix.LiveView.Plug/ProjectName.HomeLive/do_things
```

### DIFF
--- a/lib/new_relic/telemetry/phoenix.ex
+++ b/lib/new_relic/telemetry/phoenix.ex
@@ -84,12 +84,18 @@ defmodule NewRelic.Telemetry.Phoenix do
     :ignore
   end
 
-  defp phoenix_name(%{plug: controller, plug_opts: action}) when is_atom(action) do
-    "/Phoenix/#{inspect(controller)}/#{action}"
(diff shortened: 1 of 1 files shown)


goto-engineering:
We undid this for now. Can come back to it later.

PR #405, opened 2023-08-25
Add 504 errors to list of retryables

griffitr:
We were finding intermittent instances of failed 504 Gateway Timeouts when the agent first tries to connect from a cell-based architecture. This resulted in a healthy instance of our application running, but without agent reporting. Allowing retries in the case of 504s has resolved the issue for us, and we think it may be beneficial for others.
@mattbaker


mattbaker (approved):
Trialed this ourselves first, the change was successful.

mattbaker:
@griffitr I wonder if you need to rebase to pick up XX's changes to the build stuff?

PR #408, opened 2023-09-14
Aggregate dimensional metrics

XiXiaPdx:
This is a stab to build off of this PR #376 

Looking at the original comments by Vince, I think one missing piece was aggregating the dimensional metrics during the 5 second reporting interval? I took a stab at it.@tpitale , if you have some time, love to hear your thoughts on this. I know @mattbaker  is interested in getting it over the finish line. 

Also, I think that [common block ](https://github.com/newrelic/elixir_agent/commit/95bef6962b1cc9b495ad2e468a28da3cfeaadfb2#diff-acdef3942c8febdf5fe3c3d38c0f8526935fa34e02a0bac66d39560fec274362R84-R85)is accurate for `count` and `summary` - https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/report-metrics-metric-api/#optional-map-attributes.

But for `gauge`, it appears the interval.ms value isn't required. It probably does no harm 😄 As an aside, the[ old NR Telemetry SDK specs](https://source.datanerd.us/agents/telemetry-sdk-specs/blob/master/capabilities.md#gauge) also show `gauge` without interval.ms

### DIFF
--- a/lib/new_relic/harvest/telemetry_sdk/api.ex
+++ b/lib/new_relic/harvest/telemetry_sdk/api.ex
@@ -17,6 +17,14 @@ defmodule NewRelic.Harvest.TelemetrySdk.API do
     |> maybe_retry(url, payload)
   end
 
+  def dimensional_metric(metrics) do
+    url = url(:metric)
+    payload = {:metrics, metrics, generate_request_id()}
+
+    request(url, payload)
+    |> maybe_retry(url, payload)
+  end
+
   def request(url, payload) do
     post(url, payload)
   end
--- a/lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
+++ b/lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
@@ -42,7 +42,6 @@ defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
   def handle_cast(_late_msg, :completed), do: {:noreply, :completed}
 
   def handle_cast({:report, metric}, state) do
-    # TODO: merge metrics with the same type/name/attributes?
     {:noreply, %{state | metrics: merge_metric(metric, state.metrics)}}
   end
 
@@ -68,7 +67,12 @@ defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
 
     case Map.get(metrics_acc, {type, name, attributes_hash}) do
       nil ->
-        Map.put(metrics_acc, {type, name, attributes_hash}, metric)
+        case type do
+          :summary ->
+            new_summary = %{type: type, name: name, count: 1, min: new_value, max: new_value, sum: new_value, attributes: attributes}
+            Map.put(metrics_acc, {type, name, attributes_hash}, new_summary)
+          _ -> Map.put(metrics_acc, {type, name, attributes_hash}, metric)
+        end
 
       %{type: :count, name: name, value: current_value, attributes: attributes} ->
         updated_metric = %{
@@ -84,8 +88,8 @@ defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
         updated_metric = %{type: :gauge, name: name, value: new_value, attributes: attributes}
         Map.put(metrics_acc, {type, name, attributes_hash}, updated_metric)
 
-      %{type: :summary, name: _name, value: _current_value, attributes: _attributes} ->
-        # TODO
+      %{type: type, name: name, count: current_value, min: min, max: max, sum: sum, attributes: attributes} =
+       #TODO aggregate summary metric
         metrics_acc
     end
   end
--- a/lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
+++ b/lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
@@ -18,7 +18,7 @@ defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
     {:ok,
      %{
        start_time_ms: System.system_time(:millisecond),
-       metrics: []
+       metrics: %{}
      }}
   end
 
@@ -43,9 +43,7 @@ defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
 
   def handle_cast({:report, metric}, state) do
     # TODO: merge metrics with the same type/name/attributes?
-    # Take phash2 of attributes and store them as a map with the
-    # key being {type, name, phash2(attributes)} => [metrics]
-    {:noreply, %{state | metrics: [metric | state.metrics]}}
+    {:noreply, %{state | metrics: merge_metric(metric, state.metrics)}}
   end
 
   # do not resend metrics when harvest has already been reported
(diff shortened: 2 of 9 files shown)


tpitale:
I _have_ signed the CLA 👎🏼

XiXiaPdx:
I was able to send `count` and `gauge` to NR. https://staging.onenr.io/0oqQalyXGj1

But I'm having issues with `summary` and it is not clear to me why. I started a thread in help-metric

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,147 @@
...
+    {:reply, build_dimensional_metric_data(state.metrics, state), state}
+  end
+
+  # Helpers
+
+  defp merge_metric(
+         %{type: type, name: name, value: new_value, attributes: attributes} = metric,
+         metrics_acc
+       ) do
+    attributes_hash = :erlang.phash2(attributes)
```
tpitale, line 66: 💯

tpitale (approved):
Great stuff, thanks for picking this up!

XiXiaPdx:
confirmed `count` `gauge` and `summary` are all making it to NR. https://staging.onenr.io/0kERzx3KAQr

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,165 @@
...
+    {:reply, build_dimensional_metric_data(state.metrics, state), state}
+  end
+
+  # Helpers
+
+  defp merge_metric(
+         %{type: :count, name: name, value: new_value, attributes: attributes} = metric,
+         metrics_acc
+       ) do
+    attributes_hash = :erlang.phash2(attributes)
```
mattbaker, line 66: Couple thoughts

1. I don't think we _have_ to hash attributes do we?
2. If we do, I have mixed feelings about phash2. We've had to remove it various parts of our codebase already because of collisions. Maybe we do something like take the sha instead.

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,165 @@
...
+       start_time_ms: System.system_time(:millisecond),
+       metrics: %{}
+     }}
+  end
+
+  # API
+
+  @spec report_dimensional_metric(:count | :gauge | :summary, atom() | binary(), any, map()) ::
+          :ok
+  def report_dimensional_metric(type, name, value, attributes) when type in @valid_types do
```
mattbaker, line 29: I've been burned more than once by the fact that attribute values can't be null in the dimensional metric API.

> _ Values can be strings, JSON numbers, or booleans_

I'm wondering if we should scan attribute values for nil and return `{:error, ...}` if we detect it.

The metric slice will get dropped and users won't know the issue is there unless they check the `NrIntegrationEvent`

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,165 @@
...
+
+  # Helpers
+
+  defp merge_metric(
+         %{type: :count, name: name, value: new_value, attributes: attributes} = metric,
+         metrics_acc
+       ) do
+    attributes_hash = :erlang.phash2(attributes)
+
+    case Map.get(metrics_acc, {:count, name, attributes_hash}) do
```
mattbaker, line 68: You might be able to tighten up some of the merge functions with [Map.update/4](https://hexdocs.pm/elixir/1.12.3/Map.html#update/4), try it out and see how it feels. It might simplify the Map.get/Map.put pattern.

lib/new_relic/harvest/telemetry_sdk/api.ex
```diff
@@ -17,6 +17,14 @@ defmodule NewRelic.Harvest.TelemetrySdk.API do
     |> maybe_retry(url, payload)
   end
 
+  def dimensional_metric(metrics) do
+    url = url(:metric)
+    payload = {:metrics, metrics, generate_request_id()}
+
+    request(url, payload)
```
mattbaker, line 24: We may want to review the status codes this module knows about against the status codes the dimensional metrics API responds with:

https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/report-metrics-metric-api/#response-status-codes

(I haven't, and maybe there's nothing to worry about)

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -77,14 +77,12 @@ defmodule NewRelic.Harvest.TelemetrySdk.DimensionalMetrics.Harvester do
          %{type: :count, name: name, value: new_value, attributes: attributes} = metric,
          metrics_acc
        ) do
-    attributes_hash = :erlang.phash2(attributes)
```
tpitale: This is an interesting one. My concern was the memory used to store the map of attributes in full has part of the key, thus using a hashing of the attributes.

I realize that `phash2` can have collisions, but I figured it was unlikely given the name and attributes that would be used within a harvest cycle of 10 seconds (I think that's the default).

tpitale: Spoke internally with everyone and we can just ignore me!

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,164 @@
...
+    [
+      %{
+        metrics: metrics,
+        common: common(state)
+      }
+    ]
+  end
+
+  defp common(%{start_time_ms: start_time_ms}) do
+    %{"timestamp" => start_time_ms, "interval.ms" => @interval_ms}
```
mattbaker, line 162: I'm guessing this doesn't matter, but should interval.ms be a fixed number, or the literal duration between the last harvest cycle and the current harvest?

Here we're saying it's always 5s, whereas our integration in NerdGraph actually computes the delta in case the interval is ever slightly different. Maybe the interval will never drift from 5000ms enough for this to matter

XiXiaPdx: If computed, is this what you are thinking?

```suggestion
    %{"timestamp" => start_time_ms, "interval.ms" => System.system_time(:millisecond) - start_time_ms}
```

mattbaker: We'd probably want to use monotonic time, I'll send you some code to consider

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,174 @@
...
+  end
+
+  defp merge_metric(
+         %{type: type, name: name, value: new_value, attributes: attributes} = metric,
+         metrics_acc
+       ),
+       do:
+         Map.update(metrics_acc, {type, name, attributes}, metric, &update_metric(&1, new_value))
+
+  defp update_metric(
```
mattbaker, line 107: really like the way this all shaped up!

mattbaker (commented):
Looks great. Sort of back tracked on how to handle the nil attribute thing (see comment) but otherwise this looks good to me! I see Tony already gave it the thumbs up.

We should give it a go in some prod systems to see how it goes.

Did you test the memory stuff with large maps in the key? I think that was the only remaining open question

lib/new_relic/harvest/telemetry_sdk/dimensional_metrics/harvester.ex
```diff
@@ -0,0 +1,174 @@
...
+      TelemetrySdk.DimensionalMetrics.HarvestCycle
+      |> Harvest.HarvestCycle.current_harvester()
+      |> GenServer.call(:gather_harvest)
+
+  # do not accept more report messages when harvest has already been reported
+  def handle_cast(_late_msg, :completed), do: {:noreply, :completed}
+
+  def handle_cast({:report, %{name: name, attributes: attributes} = metric}, state) do
+    if nil_attributes?(attributes) do
+      NewRelic.log(
```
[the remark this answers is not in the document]
XiXiaPdx: https://github.com/newrelic/newrelic-telemetry-sdk-specs/blob/master/validation.md#sdk-validation

I thought I saw the SDK would check for attributes being null but I'm mistaken. I think we send it through for the backend to validate and generate Nr Integration errors. 

I don't think I would log a message here. It can be confusing since we are still sending it. I would remove this entirely.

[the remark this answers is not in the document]
mattbaker: sounds good to me!

Issue #419, opened 2023-12-14
Errors not recorded since upgrading to Elixir 1.15


tmaszk:
**Describe the bug**
Elixir errors are not sent to New Relic since upgrading to Elixir 1.15

**Environment**
* Elixir 1.15.7  Erlang OTP 26
* Agent version 1.28.0

tpitale:
@tmaszk I'm going to try to start looking at this. Any other details you might have would be much appreciated!

tpitale:
I think you already fixed this change in work you did for #421 https://hexdocs.pm/elixir/1.15/changelog.html#integration-with-erlang-otp-logger

tmaszk:
I think you are correct.  I added a comment to https://github.com/newrelic/elixir_agent/pull/421 https://hexdocs.pm/elixir/1.15/changelog.html#integration-with-erlang-otp-logger about how a configuration change the user would need to make.

tpitale:
Merged #426 and #427

PR #421, opened 2024-01-02
elixir 1.15 and OTP 26 compatibility

tmaszk:
All the tests are passing.  The biggest changes were having to do with the Logger changes in 1.15 and how tasks are spawned.

Trying to address https://github.com/newrelic/elixir_agent/issues/419


lib/new_relic/transaction/erlang_trace.ex
```diff
@@ -92,6 +96,18 @@ defmodule NewRelic.Transaction.ErlangTrace do
   defp trace_proc_lib_spawn_link do
     :erlang.trace_pattern({:proc_lib, :spawn_link, :_}, [{:_, [], [{:return_trace}]}], [])
     :erlang.trace_pattern({:proc_lib, :start_link, :_}, [{:_, [], [{:return_trace}]}], [])
+
+    # Starting with elixir 1.15 tasks are spawned with Task.Supervised.
+    # See https://github.com/elixir-lang/elixir/commit/ecdf68438160928f01769b3ed76e184ad451c9fe
```
tpitale, line 101: Great find! I think this was the thing I was looking for for quite some time. Thank you!

tpitale:
@tmaszk This is fantastic, thank you. I'm gonna request one more reviewer and hopefully we can merge and release this ASAP.

tmaszk:
Thanks.  I think this PR is a good step forward, but in my local testing it isn't capturing all the errors. 

From what I can understand, not all the errors go through the Erlang logger, so we might need to harvest some more of them.  I didn't want to look into that until I knew if this was on the right track or not.

tmaszk:
I resolved the merge conflict with the recent changes to master

tpitale:
Thank you @tmaszk, sorry about that.

jmschles (approved):
This looks good to me, and thanks for the PR @tmaszk! Regarding the missing errors: is the plan to look into that as a follow-up before cutting a release?

tmaszk:
Yes, I think that would be best.  I started looking at it, but haven't had time to finish it.  I could either add it to this PR or create another one.

tpitale:
Thanks @tmaszk, appreciate the perspective. I think an additional PR would be good. Thinking we merge this, do some broader testing, and then try to spot the missing errors issue.

Really appreciate all the time you're putting into this.

.github/workflows/elixir.yml
```diff
@@ -12,12 +12,19 @@ jobs:
   unit_test:
     name: Unit Tests - Elixir ${{matrix.elixir}} / OTP ${{matrix.otp}}
     runs-on: ubuntu-20.04
+    continue-on-error: true
```
XiXiaPdx, line 15: @tmaszk  👋  Could I ask you about this setting? I'm not a Github Actions expert and AFAICT, this might mean the test run passes even if there are errors? Or maybe this behaves differently than what I'm expecting.   Could you share your reasoning for adding this?

tmaszk:
@tpitale Thanks for merging.  I've tested this more with my application and I believe it does capture all the errors IF you configure the logger with the `handle_sasl_reports` flag as documented [here](https://hexdocs.pm/logger/1.12.3/Logger.html#module-erlang-otp-integration).

So when I add this line to my project's `config.exs`, it captures the logs I would expect.
```
config :logger, handle_sasl_reports: true
```

It seems like we should either:
- add the configuration above to README.md (only when using elixir 1.15 and above)
- change the error reporter to not be dependent on the `handle_sasl_reports` option

I can help with either when I get a chance.  I'm pretty busy but would probably have some time to work on it on the weekends.

tpitale:
Looks like it is mentioned as defaulting to false here: https://hexdocs.pm/logger/1.15/Logger.html, but set to tru if we want to capture, which could be the missing errors:

> Supervisor, crash, and progress reports should be logged

tmaszk:
@tpitale FYI, I've been testing some more locally and even though it captures all the errors, setting `handle_sasl_reports: true` floods my logs with an insane number of info messages.  I don't think it is a reasonable recommendation to make to users.  I think updating the error reporter to not rely on that configuration is going to be the better path forward.

lib/new_relic/error/logger_handler.ex
```diff
@@ -1,30 +1,33 @@
...
     if NewRelic.Transaction.Sidecar.tracking?() do
       NewRelic.Error.Reporter.report_error(:transaction, report)
     else
       NewRelic.Error.Reporter.report_error(:process, report)
     end
+
+    :none
+  end
+
+  def translator(_level, :error, _timestamp, {_, %{args: _, function: _}} = metadata) do
```
tpitale, line 22: @tmaszk I looked more closely at 1.14, and 1.15.7 for this: https://hexdocs.pm/logger/1.15.7/Logger.Translator.html#callbacks

We may have gotten it slightly wrong. I think the args are `translator(_min_level, _level, :error, report_tuple)`. And it seems like `kind` is either `:error` or `:report` explicitly.

tpitale: Correction :format or :report, so :error is the level being reported. Ignore me!

.github/workflows/elixir.yml
XiXiaPdx: @tmaszk  Thanks! A nice confirmation happened as well with [this test run](https://github.com/newrelic/elixir_agent/actions/runs/7611561532/job/21114894502?pr=424).

PR #424, opened 2024-01-22
Identify Phoenix LiveView metrics

tmaszk:
Before this PR, all LiveView metrics are named for the plug and action, i.e. `WebTransaction/Phoenix/Phoenix.LiveView.Plug/index`.

This PR names those metrics after the module and action, i.e. `WebTransaction/Phoenix/PhxExampleWeb.HomeLive/index`.  This provides more insight into the details of the metric.

To properly test, `phoenix_example` is updated with the latest phoenix packages and the pages and layouts with the modern Phoenix conventions.

PR #426, opened 2024-01-24
Additional elixir 1.15 compatibility

tmaszk:
This is a follow on to PR https://github.com/newrelic/elixir_agent/pull/421 and should completely address Issue https://github.com/newrelic/elixir_agent/issues/419

The purpose of the new `NewRelic.ErrorLogger` module is to prevent Logger from displaying the MANY status messages caused by enabling `handle_sasl_reports`.

I don't like requiring the user update their `:logger` config to allow the agent to capture errors, but can't think of another way to do it.

This PR also has the nice side effect of not showing the error messages in the console when running `mix test`, because it no longer logs errors to the console during the unit tests.

@tpitale FYI.  This PR plus https://github.com/newrelic/elixir_agent/pull/424, I think it would make a nice update for Elixir users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,16 @@ config :new_relic_agent,
   httpc_request_options: [connect_timeout: 5000]
 ```
 
+#### For Elixir 1.15 and higher
+
+Due to changes in the Elixir 1.15 Logger, additional logger configuration is needed for NewRelic to capture all errors. Update your logger configuration by setting `handle_sasl_reports` to `true` and adding `NewRelic.ErrorLogger` to your logger backends.
+
+```elixir
+config :logger,
+  handle_sasl_reports: true,
+  backends: [:console, NewRelic.ErrorLogger]
+```
+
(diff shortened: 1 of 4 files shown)


tpitale:
Going to try this out in one of our apps for testing, hopefully today or tomorrow.

tmaszk:
@tpitale Sounds good.  If there are additional sasl messages that need to be silenced, it  would be easy to add them to  `ErrorLogger.translate/4`.

I originally thought about silencing all `{_, :progress}` messages, not just `[:supervisor, :application_controller]` but those two were the only messages that came up in my application.

tpitale:
This has been running great. I'm going to get one other person to take a look and then we can merge and release!

Thank you, again, so much @tmaszk

tmaszk:
@tpitale That's great news! 

I'd encourage you to consider including #427 before releasing.  I made that PR after discovering a class of errors that were not being captured.

PR #427, opened 2024-01-30
parse elixir exceptions

tmaszk:
I've been testing with the master branch and have discovered a class of errors that is causing `MetadataReporter` to throw an exception.  It occurs when the original exception occurs within an async Task.   This PR contains the fix and a unit test to exercise it.

### DIFF
--- a/lib/new_relic/error/metadata_reporter.ex
+++ b/lib/new_relic/error/metadata_reporter.ex
@@ -62,19 +62,21 @@ defmodule NewRelic.Error.MetadataReporter do
     end
   end
 
-  defp parse_reason({exception, stacktrace}) do
-    type = parse_type(exception)
+  defp parse_reason({%type{message: message} = exception, stacktrace}) do
     expected = parse_error_expected(exception)
-    message = Map.get(exception, :message)
+    type = inspect(type)
     reason = "(#{type}) #{message}"
 
     {type, reason, stacktrace, expected}
   end
 
-  defp parse_type(%type{}) do
-    type
-    |> Module.split()
-    |> Enum.join(".")
+  defp parse_reason({exception, stacktrace}) do
+    exception = Exception.normalize(:error, exception, stacktrace)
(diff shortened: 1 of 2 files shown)


XiXiaPdx:
@tmaszk  I have a question about the new test. I added the test to master, expecting it to fail the assertion but it passes instead. Here is what I see in `events`.   Is that unexpected?

```
[test/error_test.exs:200: ErrorTest."test Catch a function clause error inside a Task"/1]
events #=> [
  [
    %{
      timestamp: 1707246881.171,
      type: "TransactionError",
      duration: nil,
      databaseDuration: nil,
      "error.class": "FunctionClauseError",
      "error.message": "(FunctionClauseError) no function clause matching in ErrorTest.ErrorDummy.handle_call/3",
      "error.expected": false,
      transactionName: "OtherTransaction/Elixir/ElixirProcess//ErrorTest.ErrorDummy"
    },
    %{
      process: "ErrorTest.ErrorDummy",
      stacktrace: "test/error_test.exs:11: ErrorTest.ErrorDummy.handle_call(:foo, nil, nil)\n(elixir 1.14.4) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2\n(elixir 1.14.4) lib/task/supervised.ex:34: Task.Supervised.reply/4\n(stdlib 5.1.1) proc_lib.erl:241: :proc_lib.init_p_do_apply/3\nanonymous fn() in ErrorTest.\"test Catch a function clause error inside a Task\"/1",
      test_attribute: "test_value"
    },
    %{}
  ]
]
```

tmaszk:
@XiXiaPdx Good question.  In my local development, I added the code in this PR on top of https://github.com/newrelic/elixir_agent/pull/426

If you add the test to the code in #426, the test will fail.

XiXiaPdx:
Oh, I see. Nice catch and thanks again for another contribution!

tpitale (approved):

PR #454, opened 2024-12-16
Test on latest supported versions

binaryseed:
Updates tests to run on most recent 4 Elixir versions, dropping older versions that are no longer receiving Elixir security patches

- https://hexdocs.pm/elixir/1.18.0-rc.0/compatibility-and-deprecations.html

PR #456, opened 2024-12-16
Conditional compilation to address warnings

binaryseed:
Address some warnings that arise across different Elixir versions
- `+0.0`
- `Logger.warn`

Also modify `ConditionalCompile` to not need a macro


lib/new_relic/logger.ex
```diff
@@ -81,7 +81,13 @@ defmodule NewRelic.Logger do
 
   defp elixir_logger(:debug, message), do: Logger.debug(message)
   defp elixir_logger(:info, message), do: Logger.info(message)
-  defp elixir_logger(:warn, message), do: Logger.warn(message)
+
+  if Version.match?(System.version(), ">= 1.11.0") do
```
tpitale, line 85: Since we have that `ConditionalCompile`, should this also use that?

tpitale: Also: do we support elixir before 1.11? The earliest we test against is 1.11.

binaryseed: Thanks missed this spot. Updated all of them to use `ConditionalCompile`

binaryseed: I don't know what we "support"

https://github.com/newrelic/elixir_agent/pull/454 changes the test matrix to the last 4 versions, and Elixir itself seems to support the last 5 versions with security patches.

https://hexdocs.pm/elixir/1.18.0-rc.0/compatibility-and-deprecations.html

We could adopt that, which would put us at `1.14` and above

binaryseed: Regardless, 1.10 and 1.11 are very old, so I've removed the conditionals that matched those version

binaryseed:
This one is ready for review

PR #458, opened 2024-12-16
Handle binary attribute values better

### LINKED ISSUE #400: Bad attribute on redis encoded string
**Describe the bug**

When calling `SET <string key> <encoded value>` the agent reports this as a "Bad attribute". We're using a library that calls `:erlang.binary_to_term` on the value before putting it into redis.

**Environment**
* Elixir & Erlang version (`elixir -v`): Erlang 24.0.6, Elixir 1.12.3
* Agent version (`mix deps | grep new_relic_agent`): 1.27.8

binaryseed:
We can write binary (but not string) attribute values as "[BINARY_VALUE]" instead of "[BAD_VALUE]", and not log them, since they could occur from builtin instrumentation.

Closes #400

### DIFF
--- a/lib/new_relic/util.ex
+++ b/lib/new_relic/util.ex
@@ -78,7 +78,7 @@ defmodule NewRelic.Util do
       {key, value} when is_bitstring(value) ->
         case String.valid?(value) do
           true -> [{key, value}]
-          false -> [bad_value(key, value)]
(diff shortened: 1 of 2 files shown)


tpitale (approved):
Thank you! I've been meaning to do something about this.

PR #463, opened 2024-12-18
Oban Instrumentation

binaryseed:
This PR adds automatic `Oban` instrumentation..
* Start a new `OtherTransaction` for each Oban job performed
* Start a new Distributed Trace for each Oban job performed


examples/apps/oban_example/lib/oban_example/worker.ex
```diff
@@ -0,0 +1,13 @@
+defmodule ObanExample.Worker do
+  use Oban.Worker
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"error" => message}}) do
+    {:error, message}
+  end
+
+  def perform(%Oban.Job{args: args}) do
```
binaryseed, line 9:
```suggestion
  def perform(%Oban.Job{args: _args}) do
```

PR #466, opened 2024-12-20
Use logger primary filter to hook into error reporting

binaryseed:
This PR cleans up the error reporting situation by leveraging `:logger` filters to hook into error reports. Support for these goes back to OTP 21, so we're good on all our supported versions

- https://www.erlang.org/doc/apps/kernel/logger.html#add_primary_filter/2

This also gets us back automatic error instrumentation! I've included a warning to remove the `ErrorLogger` from `:backend` configuration for anyone who has done that.

This technique is already in use inside the agent for the Logs in Context feature.


binaryseed:
Rebased & ready for review

tpitale (approved):
🎉

Issue #388, opened 2022-07-27
Provide Notice Errors in Elixir Agent


JamieMaclean:
I'd just like to say thank you for maintaining this library! So far my experience with it has been great!
 
**Is your feature request related to a problem? Please describe.**
From what I can see, error reporting is handled via cowboy telemetry events which means that handled errors are never sent to New Relic to be seen in the UI. 

I've seen that some of the other agents support this via e.g. in Python:

```
newrelic.agent.notice_error(error=None, attributes={}, expected=None, ignore=None, status_code=None, application=None)
```

It would be good to have similar functionality in Elixir.

**Describe the solution you'd like**
Creation of something similar to:

```
NewRelic.report_notice_error(%{error: nil, attributes=%{}, expected: nil, ignore: nil, status_code: nil, application: nil})
```

**Describe alternatives you've considered**
So far our work around is to send a custom event with the stacktrace and create an alert in the UI. But this is a non-optimal solution.

ivanfrias:
hi @JamieMaclean. May I ask you how have you implemented the workaround you described using this elixir agent ? Thx!

JamieMaclean:
> hi @JamieMaclean. May I ask you how have you implemented the workaround you described using this elixir agent ? Thx!

Hey @ivanfrias, sure thing. It is something along the lines of:

```elixir
try do
  raise("Raise for some reason")
rescue
  e ->
    NewRelic.report_custom_event("MyHandledError", %{
      "exception" => inspect(e),
      stacktrace: inspect(__STACKTRACE__)
    })

  continue_as_if_nothing_ever_happened()
end
```

Then, in the UI, you can create alerts based on this custom event.

Like I said, not ideal... But it works.

ivanovv:
I think Sentry has this one implemented, but they have to parse the log to achieve this. It looks like BEAM is that easy to instrument to catch all unhandled exceptions

Issue #347, opened 2021-05-11
Explicit way of failing a "other" transaction


VitorTrin:
**Is your feature request related to a problem? Please describe.**
The `NewRelic.start_transaction()` automaticaly catches exceptions and reports errors, which is great for most use cases, but sometimes you can't raise and you still would like to report something. For example, inside Oban jobs, raises are caught by Oban and never end up reported.

**Describe the solution you'd like**
Add the `NewRelic.stop_transaction/1` function. The argument of this function is the error that will be reported

**Describe alternatives you've considered**
Directly calling `NewRelic.Transaction.Reporter.fail/1` with varied levels of success.

hdtafur:
@VitorTrin I'm curious, is this not accomplishable by decorating the Oban perform function with something like `@trace :perform`. Unless I'm misunderstanding how the trace macro works, it does seem to raise the exception after creating a span for it.
https://github.com/newrelic/elixir_agent/blob/f109cedd41ab9a8b14f65fb42d36a01466511da1/lib/new_relic/tracer/macro.ex#L194-L199

VitorTrin:
The trace macro wasn't enough for me because I wanted my job transactions to show up as `Oban` transactions, not `Other`

hdtafur:
Do they show up in any distinguishable manner? The tracer documentation indicates that it reports the function name. 
https://github.com/newrelic/elixir_agent/blob/f109cedd41ab9a8b14f65fb42d36a01466511da1/lib/new_relic/tracer.ex#L23-L27
I'm invested in this as well since it is something I am also trying to get working with Oban

PR #470, opened 2024-12-28
Add notice_error function for rescued exceptions

binaryseed:
Provides a new function `NewRelic.notice_error/2` to report a rescued exception inside a Transaction. Requires passing in the exception and the `__STACKTRACE__`

Closes https://github.com/newrelic/elixir_agent/issues/388
Closes https://github.com/newrelic/elixir_agent/issues/347


tpitale (approved):
Thank you! I've been meaning to add this for quite a while.

Issue #327, opened 2021-03-11
No transactions for Absinthe


cshanes:
**Describe the bug**
After setting up the new_relix_elixir agent and new_relic_absinthe agent, the only data that shows up in the New Relic dashboard are database queries. No transactions.

I did add it to the middleware like so:
```
  def middleware(middleware, _field, %Absinthe.Type.Object{
        identifier: identifier
      })
      when identifier in [:query, :subscription, :mutation] do
    [NewRelic.Absinthe.Middleware, Instrument, RequestLogger] ++ middleware
  end
  ```

**Environment**
* Elixir & Erlang version:  Erlang/OTP 23, Elixir 1.11.3
* Agent version: new_relic_agent 1.23.6, new_relic_absinthe 0.0.4
* Absinthe version: absinthe 1.5.5

Am I missing additional configuration?

**Comment by cshanes** [comment]
Just to add a bit more information, I spent time adding debug output to the New Relic middleware and confirmed the `complete` function (https://github.com/binaryseed/new_relic_absinthe/blob/master/lib/new_relic/absinthe/middleware.ex#L42) is getting called. I printed out some of the values and everything looks fine as far as I can tell:

```
span: 
{["roundSubmissionPerformance", "roundDailyPerformances"],
 #Reference<0.2859887835.2323906562.39014>}
primary name: APITournament.Models.UserProfile.round_daily_performances_resolver
secondary name: APITournament.Schemata.roundSubmissionPerformance.roundDailyPerformances
```

tpitale:
It looks like we need to add Telemetry support for Absinthe maybe?

binaryseed:
I started that a while back, the branch is here: https://github.com/binaryseed/elixir_agent/commits/absinthe-telemetry I never got around to finishing it...

It requires an extension of the span reporting API that's not quite so simple. Spans via Absinthe don't have a reliable nested "parenting" relationship and that's what the automatic `@trace` based span generation relies on. Parenting has to be tracked manually using the `id` emitted in the telemetry events

iscottming:
Hi, @binaryseed , I found out that this branch has been held for a while, and we recently wanted to change the middleware-based approach to telemetry-based because the middleware-based new relic fails when the middleware pipeline breaks.

What I want to know now is, what else do I need to do to continue what you did before?

For example, which part do I specifically need to extend for extending the API for span reporting? What files and logic can I refer to?

[...]

PR #469, opened 2024-12-28
Add finch instrumentation

binaryseed:
Adds automatic `Finch` instrumentation based on `:telemetry` events.

Note:
* If your app already manually instruments Finch requests with `@trace`, you'll see a warning to remove that instrumentation or to disable automatic Finch instrumentation

PR #471, opened 2024-12-31
Absinthe instrumentation

binaryseed:
This PR provides automatic Absinthe instrumentation via `:telemetry`

There's an existing package https://github.com/binaryseed/new_relic_absinthe that provides absinthe instrumentation, before there was `:telemetry` instrumentation built in. That package will be retired. Users of that package should remove it in favor of this instrumentation.

Closes https://github.com/newrelic/elixir_agent/issues/327


binaryseed:
Will need to rebase after #469

PR #478, opened 2025-01-09
Add add_span_attributes function

binaryseed:
* Adds `add_span_attributes ` for adding additional attributes to the current span
* Adds `spec`s for all public API functions


tpitale (dismissed):
Nice. I like this.

PR #486, opened 2025-02-01
Fix test config

TylerWitt:
This only changes tests--

I noticed that there are many cases where tests effect Application config, and do no reset them. This can cause test flakiness, especially with `automatic_attributes` being set to `[]` after some tests, when other tests rely on the values set in `test.exs`.

These changes add a helper that resets Application config `on_exit`. The helper removes the config if it didn't previously exist, and will set it to it's original value if it did exist.

This shouldn't have any negative effects at all, unless other tests were relying on broken state to begin with.

I also fixed a type system warning emitted in Elixir 1.18.

### DIFF
--- a/test/backoff_sampler_test.exs
+++ b/test/backoff_sampler_test.exs
@@ -98,7 +98,12 @@ defmodule BackoffSamplerTest do
   end
 
   test "cycles trigger" do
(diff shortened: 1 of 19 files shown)


binaryseed:
Thanks! I included and extended this in https://github.com/newrelic/elixir_agent/pull/487

PR #487, opened 2025-02-01
Improve test configuration to reduce test flakes

binaryseed:
This PR extends #486, adopting the same pattern for setting application config in tests that we already have for the other agent configuration mechanisms. Also upgrades all of those to automatically wire up an `on_exit` function to revert their changes.

Thanks @TylerWitt for getting this started!

Closes https://github.com/newrelic/elixir_agent/pull/486


TylerWitt:
The `run_with` reads nicely, great call.

tpitale (approved):

PR #492, opened 2025-02-04
Add transaction.name to spansaction

### LINKED ISSUE #490: transaction.name missing from Span events
transaction.name is missing from Span events
https://docs.newrelic.com/attribute-dictionary/?event=Span

This attribute is required to correlate transaction metrics and names with Span events. While one span event for each transaction has _part_ of the transaction name - for example:
WebTransaction/AbsinthePlug/query/actor.account.results vs AbsinthePlug/query/actor.account.results

But since it is not an exact match one would need to know what part of the transaction name is on the Span to query it. A use case is to find distributed traces that match a given transaction metric name. But any other use cases requiring this correlation will likewise not function properly without the attribute.

binaryseed:
Adds `transaction.name` attribute to the "spans action" span event so it can be correlated to transaction metrics

Closes #490

### DIFF
--- a/examples/apps/ecto_example/test/ecto_example_test.exs
+++ b/examples/apps/ecto_example/test/ecto_example_test.exs
@@ -33,19 +33,19 @@ defmodule EctoExampleTest do
 
     assert TestSupport.find_metric(
              metrics,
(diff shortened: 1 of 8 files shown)


tpitale (approved):
Thank you!

PR #493, opened 2025-02-07
Refactor Transaction.Complete.extract_transaction_info

TylerWitt:
This was iterating through the same list multiple times when a single pass would do.

This should be more performant.

### DIFF
--- a/lib/new_relic/transaction/complete.ex
+++ b/lib/new_relic/transaction/complete.ex
@@ -108,30 +108,19 @@ defmodule NewRelic.Transaction.Complete do
 
     function_segments =
       function_segments
-      |> Enum.map(&transform_time_attrs/1)
-      |> Enum.map(&transform_trace_time_attrs(&1, tx_attrs.start_time))
-      |> Enum.map(&transform_trace_name_attrs/1)
-      |> Enum.map(&struct(Transaction.Trace.Segment, &1))
+      |> transform_function_segments(tx_attrs.start_time)
       |> Enum.group_by(& &1.pid)
       |> Enum.into(%{}, &generate_segment_tree(&1))
 
     root_process_segment =
       tx_attrs
-      |> Map.take([:name, :pid, :start_time, :end_time])
-      |> List.wrap()
-      |> Enum.map(&transform_trace_time_attrs(&1, tx_attrs.start_time))
-      |> Enum.map(&transform_trace_name_attrs/1)
-      |> Enum.map(&struct(Transaction.Trace.Segment, &1))
-      |> List.first()
+      |> transform_root_process_segment()
       |> Map.put(:id, pid)
 
     process_segments =
       process_spawns
       |> collect_process_segments(process_exits)
-      |> Enum.map(&transform_trace_time_attrs(&1, tx_attrs.start_time))
-      |> Enum.map(&transform_trace_name_attrs/1)
-      |> Enum.map(&struct(Transaction.Trace.Segment, &1))
-      |> Enum.reject(&(&1.relative_start_time == &1.relative_end_time))
+      |> transform_process_segments(tx_attrs.start_time)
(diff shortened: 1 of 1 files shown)


lib/new_relic/transaction/complete.ex
```diff
@@ -170,6 +159,40 @@ defmodule NewRelic.Transaction.Complete do
...
+      struct(Transaction.Trace.Segment, segment)
+    end
+  end
+
+  defp transform_function_segment(segment, start_time) do
+    segment
+    |> transform_time_attrs()
+    |> transform_trace_time_attrs(start_time)
+    |> transform_trace_name_attrs()
+  end
```
binaryseed, line 184: No need for another intermediate function here...

```suggestion
  defp transform_function_segments(function_segments, start_time) do
    for segment <- function_segments do
      segment =
        segment
        |> transform_time_attrs()
        |> transform_trace_time_attrs(start_time)
        |> transform_trace_name_attrs()

      struct(Transaction.Trace.Segment, segment)
    end
  end
```

lib/new_relic/transaction/complete.ex
```diff
@@ -170,6 +159,40 @@ defmodule NewRelic.Transaction.Complete do
...
+
+  defp transform_process_segments(process_segments, start_time) do
+    for segment <- process_segments,
+        time_attr_segment = transform_trace_time_attrs(segment, start_time),
+        time_attr_segment.relative_start_time != time_attr_segment.relative_end_time do
+      segment = transform_trace_name_attrs(time_attr_segment)
+
+      struct(Transaction.Trace.Segment, segment)
+    end
+  end
```
binaryseed, line 194: I'd prefer for all these functions to have a matching structure so that they can be understood & compared easily.

Also, let's pull out the `Enum.reject` part back to where it was, since it's not transforming the segments, but filtering them.

```suggestion
  defp transform_process_segments(process_segments, start_time) do
    for segment <- process_segments do
      segment =
        segment
        |> transform_trace_time_attrs(start_time)
        |> transform_trace_name_attrs(time_attr_segment)

      struct(Transaction.Trace.Segment, segment)
    end
  end
```

binaryseed (changes requested):
This is a sensitive area of the code that is fairly hard to test, because it's only after the New Relic data pipeline processes the results of this output that we know if it's working as expected.

So I'm prioritizing organizing this code in a way that is uniform and easy to compare that we're doing the same work as before...

lib/new_relic/transaction/complete.ex
TylerWitt: The reject logic here is included in the comprehension, which is more performant than another Enum iteration. If the function naming is the issue, I can change it to filter_map_process_segments or something.

binaryseed:
I got curious to see if the proposed changes made a noticeable difference, so I ran some benchmarks...

I grabbed an example from a real world app with a lot of segments and compared the existing code with the proposed changes. The results were mixed, sometimes the existing code was faster, sometimes the new code was faster. For a Transaction with 200 traced segments, times were in the ballpark of 280 μs, and the differences were generally within 5 μs.

Given the inconclusive results, and the fact that this code is executed asynchronously from the user's application code in a separate Sidecar process, _and_ given that the end result of the `Complete.run` function is to put data into harvester queues where it will just wait for the next cycle to send data up to new relic (5 seconds or 60 seconds depending on the data type), I'm going close this PR, and keep things how they are...

TylerWitt:
> I grabbed an example from a real world app with a lot of segments and compared the existing code with the proposed changes. The results were mixed, sometimes the existing code was faster, sometimes the new code was faster. For a Transaction with 200 traced segments, times were in the ballpark of 280 μs, and the differences were generally within 5 μs.

Is there a way I can test things locally like this without a license? Or no?

Also thanks for the testing! Interesting it wasn't a bigger difference.

PR #494, opened 2025-02-07
Support native JSON library

TylerWitt:
Jason has been sunset for later versions of Elixir, and the new JSON library is available as of 1.18.

This could allow less dependencies for apps.

### DIFF
--- a/lib/new_relic/harvest/collector/protocol.ex
+++ b/lib/new_relic/harvest/collector/protocol.ex
@@ -93,9 +93,10 @@ defmodule NewRelic.Harvest.Collector.Protocol do
 
   defp parse_http_response({:ok, %{status_code: 200, body: body}}, _params) do
     {:ok, NewRelic.JSON.decode!(body)}
-  rescue error ->
-    NewRelic.log(:error, "Bad collector JSON: #{Exception.message(error)}")
-    {:error, :bad_collector_response}
+  rescue
+    error ->
+      NewRelic.log(:error, "Bad collector JSON: #{Exception.message(error)}")
+      {:error, :bad_collector_response}
   end
 
   defp parse_http_response({:ok, %{status_code: 202}}, _params) do
--- a/lib/new_relic/harvest/collector/protocol.ex
+++ b/lib/new_relic/harvest/collector/protocol.ex
@@ -92,14 +92,10 @@ defmodule NewRelic.Harvest.Collector.Protocol do
   end
 
   defp parse_http_response({:ok, %{status_code: 200, body: body}}, _params) do
-    case Jason.decode(body) do
-      {:ok, response} ->
-        {:ok, response}
-
-      {:error, jason_exception} ->
-        NewRelic.log(:error, "Bad collector JSON: #{Exception.message(jason_exception)}")
-        {:error, :bad_collector_response}
-    end
+    {:ok, NewRelic.JSON.decode!(body)}
+  rescue e ->
+    NewRelic.log(:error, "Bad collector JSON: #{Exception.format_banner(:error, e)}")
+    {:error, :bad_collector_response}
   end
 
   defp parse_http_response({:ok, %{status_code: 202}}, _params) do
@@ -179,7 +175,7 @@ defmodule NewRelic.Harvest.Collector.Protocol do
   end
 
   defp log_error(status, error, params, body) do
-    case Jason.decode(body) do
+    case NewRelic.JSON.decode(body) do
       {:ok, %{"exception" => exception}} ->
         NewRelic.log(
           :error,
--- a/lib/new_relic/harvest/collector/protocol.ex
+++ b/lib/new_relic/harvest/collector/protocol.ex
@@ -93,8 +93,8 @@ defmodule NewRelic.Harvest.Collector.Protocol do
 
   defp parse_http_response({:ok, %{status_code: 200, body: body}}, _params) do
     {:ok, NewRelic.JSON.decode!(body)}
-  rescue e ->
-    NewRelic.log(:error, "Bad collector JSON: #{Exception.format_banner(:error, e)}")
+  rescue error ->
+    NewRelic.log(:error, "Bad collector JSON: #{Exception.message(error)}")
     {:error, :bad_collector_response}
   end
 
--- a/lib/new_relic/json.ex
+++ b/lib/new_relic/json.ex
@@ -15,5 +15,4 @@ defmodule NewRelic.JSON do
     true ->
       raise "[:new_relic_agent] No JSON library found, please add :jason as a dependency"
   end
-
 end
--- a/lib/new_relic/json.ex
+++ b/lib/new_relic/json.ex
@@ -1,15 +1,19 @@
 defmodule NewRelic.JSON do
   @moduledoc false
 
-  if Code.ensure_loaded?(JSON) do
-    defdelegate decode(data), to: JSON
-    defdelegate decode!(data), to: JSON
-    defdelegate encode!(data), to: JSON
+  cond do
+    Code.ensure_loaded?(JSON) ->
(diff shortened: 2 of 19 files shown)


lib/new_relic/json.ex
```diff
@@ -0,0 +1,15 @@
...
+    defdelegate decode(data), to: JSON
+    defdelegate decode!(data), to: JSON
+    defdelegate encode!(data), to: JSON
+  end
+
+  if Code.ensure_loaded?(Jason) do
+    defdelegate decode(data), to: Jason
+    defdelegate decode!(data), to: Jason
+    defdelegate encode!(data), to: Jason
+  end
```
binaryseed, line 14: I think there will be many situations where both modules are loaded. That means we'd execute both of these blocks and call `defdelegate` twice...

It will be a long time before `Jason` actually drops off as a dependency, if ever. If it's a dependency for anything in an app, it'll be there.

* We should use a `cond` here to prioritize defining the functions
* We should not reference the module directly with defdelegate, or it'll warn about missing modules
* We should have a fallback that prompts the user

```suggestion
  cond do
    Code.ensure_loaded?(JSON) ->
      def decode(data), do: apply(JSON, :decode, [data])
      def decode!(data), do: apply(JSON, :decode!, [data])
      def encode!(data), do: apply(JSON, :encode!, [data])

    Code.ensure_loaded?(Jason) ->
      def decode(data), do: apply(Jason, :decode, [data])
      def decode!(data), do: apply(Jason, :decode!, [data])
      def encode!(data), do: apply(Jason, :encode!, [data])

    true ->
      raise "[:new_relic_agent] No JSON library found, please add :jason as a dependency"
  end

```

lib/new_relic/util/http.ex
```diff
@@ -14,14 +14,11 @@ defmodule NewRelic.Util.HTTP do
...
-
-      {:error, message} ->
-        NewRelic.log(:debug, "Unable to JSON encode: #{inspect(body)}")
-        {:error, message}
-    end
+    body = NewRelic.JSON.encode!(body)
+    post(url, body, headers)
+  rescue e ->
+    NewRelic.log(:debug, "Unable to JSON encode: #{inspect(body)}")
+    {:error, Exception.format_banner(:error, e, __STACKTRACE__)}
```
binaryseed, line 21: Will you please unwind this and add a `NewRelic.JSON.encode/1` function, I prefer not to use exceptions as control flow unless absolutely necessary...

lib/new_relic/harvest/collector/protocol.ex
```diff
@@ -92,14 +92,10 @@ defmodule NewRelic.Harvest.Collector.Protocol do
...
-        {:ok, response}
-
-      {:error, jason_exception} ->
-        NewRelic.log(:error, "Bad collector JSON: #{Exception.message(jason_exception)}")
-        {:error, :bad_collector_response}
-    end
+    {:ok, NewRelic.JSON.decode!(body)}
+  rescue e ->
+    NewRelic.log(:error, "Bad collector JSON: #{Exception.format_banner(:error, e)}")
+    {:error, :bad_collector_response}
```
binaryseed, line 98: Same here, use `decode/1`...

binaryseed (changes requested):

lib/new_relic/util/http.ex
TylerWitt: Ah, the built in JSON does not have an `encode/1` https://hexdocs.pm/elixir/JSON.html#encode!/2

I can't find the reasoning off the top of my head, but from what I remember the argument was that `encode` would raise in most cases there were errors anyways, or something along those lines.

lib/new_relic/harvest/collector/protocol.ex
TylerWitt: The reason I switched here was that the error tuple that is returned from the native version was less descriptive, like `{:invalid_sequence, 60}` or something like that, and the prior handling in the `NewRelic.log` errored.

To keep the same logging, this is the way to go. Not sure what is preferred here.

lib/new_relic/util/http.ex
binaryseed: Interesting. Only request is that we trim down the error message since in some places it'll get embedded inside a log message with other content

```suggestion
    body = NewRelic.JSON.encode!(body)
    post(url, body, headers)
  rescue error ->
    NewRelic.log(:debug, "Unable to JSON encode: #{inspect(body)}")
    {:error, Exception.message(error)}
```

lib/new_relic/harvest/collector/protocol.ex
binaryseed:
```suggestion
    {:ok, NewRelic.JSON.decode!(body)}
  rescue error ->
    NewRelic.log(:error, "Bad collector JSON: #{Exception.message(error)}")
    {:error, :bad_collector_response}
```

binaryseed (changes requested):

lib/new_relic/util/http.ex
```diff
@@ -14,14 +14,11 @@ defmodule NewRelic.Util.HTTP do
...
-
-      {:error, message} ->
-        NewRelic.log(:debug, "Unable to JSON encode: #{inspect(body)}")
-        {:error, message}
-    end
+    body = NewRelic.JSON.encode!(body)
+    post(url, body, headers)
+  rescue error ->
+    NewRelic.log(:debug, "Unable to JSON encode: #{inspect(body)}")
+    {:error, Exception.message(error)}
```
binaryseed, line 21:
```suggestion
  rescue
    error ->
      NewRelic.log(:debug, "Unable to JSON encode: #{inspect(body)}")
      {:error, Exception.message(error)}
```

lib/new_relic/harvest/collector/protocol.ex
```diff
@@ -92,14 +92,10 @@ defmodule NewRelic.Harvest.Collector.Protocol do
...
-        {:ok, response}
-
-      {:error, jason_exception} ->
-        NewRelic.log(:error, "Bad collector JSON: #{Exception.message(jason_exception)}")
-        {:error, :bad_collector_response}
-    end
+    {:ok, NewRelic.JSON.decode!(body)}
+  rescue error ->
+    NewRelic.log(:error, "Bad collector JSON: #{Exception.message(error)}")
+    {:error, :bad_collector_response}
```
binaryseed, line 98:
```suggestion
  rescue
    error ->
      NewRelic.log(:error, "Bad collector JSON: #{Exception.message(error)}")
      {:error, :bad_collector_response}
```

binaryseed (changes requested):

lib/new_relic/json.ex
```diff
@@ -0,0 +1,19 @@
...
+    Code.ensure_loaded?(Jason) ->
+      def decode(data), do: apply(Jason, :decode, [data])
+      def decode!(data), do: apply(Jason, :decode!, [data])
+      def encode!(data), do: apply(Jason, :encode!, [data])
+
+    true ->
+      raise "[:new_relic_agent] No JSON library found, please add :jason as a dependency"
+  end
+
+end
```
binaryseed, line 19:
```suggestion
  end
end
```

lib/new_relic/util/vendor.ex
```diff
@@ -58,7 +58,7 @@ defmodule NewRelic.Util.Vendor do
   def aws_vendor_map(url) do
     case :httpc.request(:get, {~c(#{url}), []}, [{:timeout, 100}], []) do
       {:ok, {{_, 200, ~c"OK"}, _headers, body}} ->
-        case Jason.decode(body) do
+        case NewRelic.JSON.decode(body) do
```
binaryseed, line 61:
```suggestion
        case NewRelic.JSON.decode(to_string(body)) do
```

binaryseed (approved):

PR #506, opened 2025-02-23
Handle report messages in logs_in_context

binaryseed:
This PR extends the Logs in Context feature to handle log messages that are "reports", ie: they don't have a string message, just a keyword list / map. For example:

```elixir
Logger.error(user: 123, error: "something went wrong")
```


tpitale (approved):
Thank you for this!

PR #510, opened 2025-02-25
Fix elixir_logger/2

thiagogsr:
There isn't a pattern matching for the `:warning` atom. It's called by `log/2` which has a guard clause to make sure the level is one of `[:debug, :info, :warning, :error]`, so `:warn` is not a valid option. I also checked who calls `NewRelic.log/2` which delegates to this function, and there isn't occurrences of `:warn`, only `:warning`.

### DIFF
--- a/lib/new_relic/logger.ex
+++ b/lib/new_relic/logger.ex
@@ -82,7 +82,7 @@ defmodule NewRelic.Logger do
 
   defp elixir_logger(:debug, message), do: Logger.debug(message)
   defp elixir_logger(:info, message), do: Logger.info(message)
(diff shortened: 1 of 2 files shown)


binaryseed:
Thanks, good find.

This could be tested easily by adding a line like `NewRelic.log(:warning, "OOPS")` to https://github.com/newrelic/elixir_agent/blob/master/test/logger_test.exs#L38

Also, you'll have to accept the CLA for this PR to be merged...

thiagogsr:
> This could be tested

Done.

binaryseed (approved):

PR #513, opened 2025-02-27
Fix Finch attributes

binaryseed:
This PR fixes Finch instrumentation `http` attributes so they are the correct ones to light up parts of the Distributed Tracing UI

### DIFF
--- a/lib/new_relic/telemetry/finch.ex
+++ b/lib/new_relic/telemetry/finch.ex
@@ -130,9 +130,9 @@ defmodule NewRelic.Telemetry.Finch do
           category: "http",
           attributes:
             %{
-              "request.url": url,
-              "request.method": request.method,
-              "request.client": "Finch",
+              url: url,
+              method: request.method,
+              component: "Finch",
               "finch.pool": finch_pool
             }
             |> Map.merge(result_attrs)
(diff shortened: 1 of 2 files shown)


lib/new_relic/telemetry/finch.ex
```diff
@@ -130,9 +130,9 @@ defmodule NewRelic.Telemetry.Finch do
           category: "http",
           attributes:
             %{
-              "request.url": url,
-              "request.method": request.method,
-              "request.client": "Finch",
+              url: url,
```
tpitale, line 133: Just out of curiosity, where did you find that the correct attributes are `url` and no longer `request.url`? I feel like I've tried looking for those docs before and failed 🙃

binaryseed: The `set_span` function has function heads for the span types that the UI knows about. `url` turns into `http.url` actually

binaryseed: This is where the knowledge of the per-"category" attributes that the UI knows about lives:

https://github.com/newrelic/elixir_agent/blob/master/lib/new_relic/span/event.ex#L70-L73

I don't know if it's documented anywhere. Also, it's always been "http.url", I just forgot that there's special attributes that are needed when I wrote the Finch instrumentation.

binaryseed: I found it: 
- https://docs.newrelic.com/docs/distributed-tracing/trace-api/trace-api-decorate-spans-attributes/#decorate

tpitale (approved):

PR #529, opened 2025-05-22
Move `:os_mon` from included app to extra app

### LINKED ISSUE #528: `included_applications: [:os_mon]` causes critical Mix error
**Describe the bug**

When running `mix release` in our application, we now get the following error:
`** (Mix) :os_mon is listed both as a regular application and as an included application`

Searching through our dependencies, `new_relic_agent` appears to be the culprit:  in the `mix.exs` there is the following line included:
`included_applications: [:os_mon]`

**Environment**
* Elixir & Erlang version (`elixir -v`): 1.18.3
* Agent version (`mix deps | grep new_relic_agent`): 1.35.0

APB9785:
## Resolves #528 

Reverts part of a previous commit https://github.com/newrelic/elixir_agent/commit/2156a34da3838e1f8e0b33103211af1144c44746 which causes Elixir/Phoenix apps to fail `mix release` with error
`** (Mix) :os_mon is listed both as a regular application and as an included application`

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -22,8 +22,7 @@ defmodule NewRelic.Mixfile do
 
   def application do
     [
-      extra_applications: [:logger, :inets, :ssl],
-      included_applications: [:os_mon],
+      extra_applications: [:logger, :inets, :ssl, :os_mon],
       mod: {NewRelic.Application, []}
     ]
   end


mix.exs
```diff
@@ -22,8 +22,7 @@ defmodule NewRelic.Mixfile do
 
   def application do
     [
-      extra_applications: [:logger, :inets, :ssl],
-      included_applications: [:os_mon],
+      extra_applications: [:logger, :inets, :ssl, :os_mon],
```
binaryseed, line 25: I put it in `included_applications` because `extra_applications` are always started, and I wanted to prevent `os_mon` from starting unless the agent is actually enabled (https://github.com/newrelic/elixir_agent/pull/473).

I didn't see that there's a way to specify them as optional however, which is probably what we want:

```suggestion
      extra_applications: [:logger, :inets, :ssl, {:os_mon, :optional}],
```

https://hexdocs.pm/mix/1.18.4/Mix.Tasks.Compile.App.html

binaryseed:
Note: Updated to mark `os_mon` as optional, since we want to avoid starting it if the agent is not enabled.

PR #534, opened 2025-06-15
Handle errors in `Harvest.HarvestCycle.send_harvest/2`

### LINKED ISSUE #511: NewRelic.Harvest.Collector.SpanEvent.HarvesterSupervisor: the process is not alive
**Describe the bug**
I started receiving a lot of errors in my production environment via Sentry.
I recently added tracing to a module that wraps ExAws.S3 calls and I think this could be the issue.

Sample code:

```elixir
defmodule MyApp.AWS do
  use NewRelic.Tracer

  @trace {:put_object, category: :external}
  def put_object(args)
    NewRelic.set_span(:generic, operation: "put_object", service: "S3")

    bucket
    |> ExAws.S3.put_object(key, body)
    |> ExAws.Request!()
  end
end
```

The errors look like this:

```
exited in: GenServer.call(NewRelic.Harvest.Collector.SpanEvent.HarvesterSupervisor, {:terminate_child, #PID<0.815077.0>}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
```

However I can see the data on NewRelic so I am assuming it's working.
I would like to get rid of the errors.
Thank you!

**Environment**
* Elixir & Erlang version (`elixir -v`): 1.17.3 OTP 26
* Agent version (`mix deps | grep new_relic_agent`): 1.32.0

sgessa:
This is an attempt to silent the error from `Harvest.HarvestCycle.send_harvest/2` when the given `supervisor` is not started.

I wasn't able to reproduce this issue consistently.

Fixes #511 

## Example

```
exited in: GenServer.call(NewRelic.Harvest.Collector.CustomEvent.HarvesterSupervisor, {:terminate_child, #PID<0.1987228.0>}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
```

## Stacktrace

```
  File "lib/gen_server.ex", line 1121, in GenServer.call/3
  File "lib/new_relic/harvest/harvest_cycle.ex", line 134, in anonymous fn/2 in NewRelic.Harvest.HarvestCycle.send_harvest/2
  File "lib/task/supervised.ex", line 101, in Task.Supervised.invoke_mfa/2
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ erl_crash.dump
 tmp
 config/secret.exs
 .DS_Store
+.tool-versions
--- a/lib/new_relic/harvest/harvest_cycle.ex
+++ b/lib/new_relic/harvest/harvest_cycle.ex
@@ -131,11 +131,20 @@ defmodule NewRelic.Harvest.HarvestCycle do
           :exit, _exit ->
             NewRelic.log(:error, "Failed to send harvest from #{inspect(supervisor)}")
         after
-          DynamicSupervisor.terminate_child(supervisor, harvester)
+          terminate_child(supervisor, harvester)
         end
       end,
       shutdown: @harvest_timeout
     )
+  end
+
+  defp terminate_child(supervisor, harvester) do
+    try do
+      DynamicSupervisor.terminate_child(supervisor, harvester)
+    catch
+      :exit, _exit ->
+        :ok
+    end
   end
 
   defp stop_harvest_cycle(timer), do: timer && Process.cancel_timer(timer)


binaryseed:
When you see this, does it line up with when your app deploys? I'm wondering if it's related to the shutdown process that the agent attempts when the supervision tree is shutting down

sgessa:
Yes it does

binaryseed:
Closing in favor of https://github.com/newrelic/elixir_agent/pull/542 which does the same but avoids nested `try` blocks

PR #536, opened 2025-07-07
fix compile time regex for otp 28

hez:
In OTP 28 you can no longer have compile time regex, ie in a module attr. Their recommendation is to move them to functions.


binaryseed:
Thanks, closing in favor of https://github.com/newrelic/elixir_agent/pull/537 which includes this commit and updates the rest of the module attribute regexes.

PR #544, opened 2025-08-07
Fix incorrect spec for report_custom_metric

tgrk:
A small fix for type spec declaration incorrectly referencing `report_custom_event/2` instead of `report_custom_metric/2` (probably a copy-paste error).


binaryseed (approved):
Thanks!

PR #548, opened 2025-09-12
Safely access :request_method on bandit start events

mashton:
82a0b1e added a path in which [:bandit, :server, :start] events missing `Plug.Conn` could propagate downstream without the `:request_id` attribute.

A dot-style accessor for `:request_method` downstream causes processes to crash, reporting the error from the New Relic Agent rather than the bandit error which caused the initial event.


binaryseed (approved):
Thanks!

